PR #9, opened 2015-08-16
Develop

lantti:
I recently got up and installed rox again. I left some notes in the INSTALL file for other people who might be installing rox on slow machines and fixed one typo I noticed in an include statement.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -33,6 +33,14 @@ Download
      pushing code.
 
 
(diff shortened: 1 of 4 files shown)


lantti:
I added in another commit to provisorily "fix" (really just hide the symptoms) an issue on the country pages when a region where a city should belong is undefined and causes buggy display.

PR #95, opened 2019-01-06
Improve INSTALL.md

tjadowski:


thisismeonmounteverest:
I never needed the sudos for mysql as the server is accessible without.

The htdocs contained very old code and was dropped. If there are still places that needs it those should be adapted instead of re-adding the directory.

tjadowski:
sudo is needed because of this case https://stackoverflow.com/questions/39281594/error-1698-28000-access-denied-for-user-rootlocalhost

thisismeonmounteverest:
I guess not a lot of people will start with a new server. Adding it doesn't do any harm though.

Where did you need the htdocs?

tjadowski:
1. In original INSTALL.md was "htdocs/exception.log" that why I needed mkdir htdocs first.
2. I've just checked sudo mysql and you are right is needed of for first root password change. I'll update pull request.

crumbking:
Hello Thomasz,
well I see this as an improvement so let's merge this one.
I just run into permission issues with a fresh ubuntu installation. I would be happy to see some more contributions.

PR #113, opened 2019-10-20
Fix mapping

tjadowski:
I don't sure that it is OK, but this eliminate error after login when login_check is execute.

### DIFF
--- a/src/Entity/LoginMessagesAcknowledged.php
+++ b/src/Entity/LoginMessagesAcknowledged.php
@@ -29,7 +29,7 @@ class LoginMessagesAcknowledged
     /**
      * @var LoginMessage
      *
-     * @ORM\OneToMany(targetEntity="LoginMessage", mappedBy="id")
+     * @ORM\OneToOne(targetEntity="\App\Entity\LoginMessage")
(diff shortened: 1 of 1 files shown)


thisismeonmounteverest:
I do not have an error when login_check is executed. What problem shall this fix?

Additionally there might be more than one unread and valid login messages so changing from OneToMany to OnoToOne doesn't seem to be correct.

tjadowski:
When I log in I have an exception like this (I have a fresh instance witch follows INSTALL.md)
Screenshot from 2019-10-18 18-39-30

And DB Schema has a problem with mapping I suppose, see
Screenshot from 2019-10-21 15-49-07

And finally if I think correctly if message can be OneToMany related it should be a ArrayCollection, right? Not LoginMessage variable?

thisismeonmounteverest:
While I get the error message on schema:validate I still do not see any problem at login_check. Weird.

OneToOne is correct, I was looking at the wrong thing (LoginMessage not LoginMessagesAcknowledged).

tjadowski:
I found this messageId field in some migration file, maybe you have a DB builded from migrations not from Entity schema?

PR #140, opened 2020-05-23
RFR: Replace SwiftMailerBundle usage by symfony/mailer

vincentchalamon:


thisismeonmounteverest:
The CommentController should use the MailerTrait instead. Forcing the locale for the recipient is necessary (that you can't set the locale for a mail you want to send is rather strange) so I wrote a utility class for that.

thisismeonmounteverest:
Every pull request seems to include an update for the symfony.lock file :)

vincentchalamon:
In this case, the receiver is not a Member object, so I needed to update the MailerTrait::sendTemplateEmail method.

> Every pull request seems to include an update for the symfony.lock file :)

I don't understand what you mean here :thinking:

thisismeonmounteverest:
The MailerInterface isn't necessary on the method anymore.
The addFlash with the translator call should be replaced with addTranslatedFlash (and include the matching trait).

thisismeonmounteverest:
I like it. Still don't understand why the symfony.lock file needs to be updated as part of the pull request (it's also part of the pull request for docker support).

vincentchalamon:
The symfony.lock file includes the list of the recipes installed in this project (and some of their dependencies). As I removed symfony/swiftmailer-bundle, I needed to remove its recipe from the symfony.lock file.

About the docker PR, the symfony.lock file shouldn't have been updated, it could be rollbacked but as it's the same modifications as in the current PR, it will be the same in final.

thisismeonmounteverest:
Is there a way to test this before merging?

vincentchalamon:
I don't know, I haven't seen any Behat test in this project...

thisismeonmounteverest:
I meant a way to use the pull request locally without merging it and I agree Rox lacks tests as a whole.

vincentchalamon:
Can't you checkout my branch locally to test it?

thisismeonmounteverest:
@vincentchalamon I added a BewelcomeAddressTrait with one of the commits today, but we can use that in CommentController after the merge.
But I just added a comment and expected the mail to be sent to me@bewelcome.org but it arrived as usual in my papercut mailtrap.

thisismeonmounteverest:
@vincentchalamon Sending another member a message isn't working anymore.

vincentchalamon:
@thisismeonmounteverest I've fixed the "Contact member" mail send, using Docker the mail should be catched on http://localhost:1080.

About the "Add a comment" mail send, this is from `modules` directory which is (I guess) legacy code. The `modules/mail/lib/mail.lib.php` instanciates a Swift_Mailer object from its own configuration, which default loads _localhost_:
> Connection could not be established with host localhost :stream_socket_client(): unable to connect to localhost:25 (Address not available)

vincentchalamon:
After adding a comment on a member, when I go back to the same member profile, I get the following error:
member_comment

thisismeonmounteverest:
@vincentchalamon If I understood the difference correctly then you just removed the ... from the second merge? Won't that fail if $receiver is not an instance of Member?

Sorry can't test that at the moment.

vincentchalamon:
Yep I just removed the `...` in the array_merge, and rebases from bootstrap4 branch.

It may fail only if the developer send an invalid $receiver. Currently, all usages are an instance of Member, but from CommentController, I directly inject an Address object, which is still valid in this case.

thisismeonmounteverest:
@vincentchalamon The first address you set in the CommentController should come through BewelcomeAddressTrait. The $receiver address should be set through the same as well in MailerTrait.

Hoping that no developer ever sends an invalid $receiver is an interesting strategy, but better play it safe.

vincentchalamon:
I updated the CommentController, and added a security on the MailerTrait::sendTemplateEmail to ensure the $receiver argument is a Member object (automatically converted to an Address object), or an Address object. Otherwise, an exception should be thrown.

thisismeonmounteverest:
@vincentchalamon I tried sending massmail and forums notifications and while the database gets updated the mails aren't sent.

To test with the docker image: Create a new forum thread through http://localhost:8080/forums/new and make sure to check the box to be kept informed.

Run php bin/console send:notifications.

For the massmailing that's too difficult to describe :)

vincentchalamon:
@thisismeonmounteverest Do you know where I can configure the `lib`/`build`/`modules` directories about database and mailer?

thisismeonmounteverest:
@vincentchalamon you may need to copy the rox_default.ini file over into the docker container. If that doesn't help, I need more information about the problem you're trying to solve.

I also tried following my own advice regarding creating a new forum thread and that runs into a brick wall, as the (old) code still expects the tags to be part of the forums_threads table. That needs a thorough code update.

vincentchalamon:
I've created a rox_local.ini with the Docker informations.

When I try to post a new thread in the forum (/forum/new), I get the following error:
forum

thisismeonmounteverest:
@vincentchalamon That's the brick wall I was talking about.
I fix a few things now and then take care of the test database.

vincentchalamon:
@thisismeonmounteverest As far as I know, the rox_*.ini seems useless cause the LegacyController will init the database from doctrine (cf. LegacyController lines 55 to 61).

As you said, the forum expects some tables which don't exist (`tags_threads`, `forum_tags`, etc.). That's out of my knowledge on this project. Don't hesitate to tell me if I can do something to help.

thisismeonmounteverest:
@vincentchalamon sorry for not replying earlier. I removed a lot of old code locally and tested everything around the forums/groups and push that tonight after a final test with an existing database.

We should be able to continue with this pull request then.

thisismeonmounteverest:
@vincentchalamon How do you fix conflicts in composer.lock?

vincentchalamon:
I've choosen the version from bootstrap4 branch, and re-did the same action I've previously done:
```
composer remove symfony/swiftmailer-bundle --no-scripts
```

thisismeonmounteverest:
@vincentchalamon The test database should now allow to send messages and requests. Create new forums posts (and group posts after you went through the group creation process) and generate entries for new notifications accordingly to be sent by send:notifications.

vincentchalamon:
@thisismeonmounteverest I've rebased from bootstrap4 branch and solved conflicts and CS issues.

After creating some forum posts and executing `bin/console send:notifications`, I see 1 email sent and am able to check it on http://localhost:1080 (mailcatcher).

Feel free to test it on your side :-)

crumbking:
@vincentchalamon 
I talked with @thisismeonmounteverest yesterday about the pull request. Seems almost good to go. We thought another test with sending a newsletter would be good.

vincentchalamon:
@crumbking How can I test to send any newsletter?

thisismeonmounteverest:
@vincentchalamon I do that test tonight (too complicated to describe the process to get there).

thisismeonmounteverest:
@vincentchalamon A question regarding the dev configuration: Why didn't you follow https://symfony.com/doc/current/mailer.html#always-send-to-the-same-address ?

vincentchalamon:
My bad, I should add it in dev mode. Currently, running this project on Docker, all the emails will be catched to the mailer service (cf. https://github.com/BeWelcome/rox/blob/bootstrap4/docker-compose.yml#L55).

thisismeonmounteverest:
@vincentchalamon What about updating services.yaml?

vincentchalamon:
I must do some tests tomorrow, to ensure the mails are still sent after this last modifications.

vincentchalamon:
OK for me, could you do some tests on your side too, to ensure nothing broke with this PR, please?

thisismeonmounteverest:
@vincentchalamon Will do tonight. Test yesterday regarding newsletter and other stuff worked out fine.
I was only surprised that the mails were all delivered to me@bewelcome.org as bcc. That's why I checked the symfony documentation leading to my comment.

thisismeonmounteverest:
@vincentchalamon Did my tests and found no issues. Good to go for me.

Rebase or merge directly?

vincentchalamon:
You can use the "Squash and merge" button to merge this PR :+1:

PR #138, opened 2020-05-22
RFR: Optimize docker configuration

vincentchalamon:
### Requirements

- [Docker](https://docs.docker.com/get-docker/)
- [Docker Compose](https://docs.docker.com/compose/install/)
- Make

### Test this PR

```shell
git clone git@github.com:vincentchalamon/rox.git
cd rox
git checkout dockerfile
make install
```

:warning: If you need to run Docker Compose as sudo, replace the last line with `make install root=1`

PR #148, opened 2020-05-31
RFR: Allow geoname import on install

vincentchalamon:
- [x] Import geonames data on `make install-geonames`
- [x] Add documentation for install using Docker

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,6 +11,41 @@ using a browser or use curl _url_ > _filename_.
 
 Production OS is Debian GNU/Linux Strech.
 
+## Install using Docker and Docker Compose
+
+1. Clone via Git:
+
+    ```bash
+    $ git clone https://github.com/BeWelcome/rox.git
+    ```
+
+1. Install using Docker and Docker Compose:
+
+    ```bash
+    $ make install
+    ```
+
+   If you need to run Docker Compose as sudo, run the following:
+
+    ```bash
+    $ make install root=1
+    ```
+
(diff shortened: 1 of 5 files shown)


thisismeonmounteverest:
@vincentchalamon While that will work that will create a really huge database and slow down the creation of the containers.
Do a make install-geonames instead?

vincentchalamon:
@thisismeonmounteverest I agree, but without the geonames the project will have a lot of 500 errors (cf. #138)

thisismeonmounteverest:
@vincentchalamon If you mean the problem on the profile page that isn't realated to missing geonames data but to wrong database seeding (and a problem in the old code I believe).

vincentchalamon:
That's the problem I was talking about. And how could we fix it?

thisismeonmounteverest:
@vincentchalamon I'll fix that.
Having a way to add the full geonames to the db is usefull nevertheless I think.

vincentchalamon:
Trying to load `import.sql` into the database, but I'm stuck with the following error:
> ERROR 1005 (HY000) at line 37: Can't create table `bewelcome`.`geonamescountries` (errno: 150 "Foreign key constraint is incorrectly formed")

vincentchalamon:
@thisismeonmounteverest Ready for review :-)

Importing geonames data is quite (very) long, but it fixes the issue on the profile:
bewelcome_profile

thisismeonmounteverest:
@vincentchalamon It certainly does :)

But the reason it shows up in the first place is a problem in the fixture. And there are a few entities missing/incorrect related to this.

PR #149, opened 2020-05-31
RFR: Fix travis

vincentchalamon:


vincentchalamon:
BTW @thisismeonmounteverest, what do you think of adding GitHub Actions as CI in every PR to check for the basic tests?: phpcpd, phploc, phpmd, php-code-sniffer, phpunit

Also adding a Docker job to ensure the Docker configuration is buildable could be interesting, but it might slow down the CI...

thisismeonmounteverest:
@vincentchalamon Everything that makes the builds more stable will bewelcome.

PR #150, opened 2020-05-31
Add quick start guide using docker

guysoft:
Result of #138


thisismeonmounteverest:
@guysoft I agree with @vincentchalamon here (see [comment ](https://github.com/BeWelcome/rox/pull/138#issuecomment-636488621) on #138).

guysoft:
Ops, missed it because I didn't look 4 hours ago :)

PR #151, opened 2020-05-31
Fix travis build

vincentchalamon:
Ping @thisismeonmounteverest


thisismeonmounteverest:
@vincentchalamon Another one? The build will still fail as I did a really bad copy&paste job...

vincentchalamon:
The previous fix was wrong because the syntax was creating 2 jobs: 1 per environment variable. But we need them both together

PR #165, opened 2020-06-15
fix blur photo

### LINKED ISSUE #164: Photo gallery on profile
Hey,
When I check a photo from the gallery on user profile, the photo shows a bad resolution in full screen.
My guess is that it is the thumbnail that is being show instead of high res version of it.

Reported here: https://www.bewelcome.org/forums/s20975/

Tr4sK:
Fix #164
I just used grep to find the issue and tested it via the docker install. Someone has to check if it does not break something else.


crumbking:
@Tr4sK 
Thanks for your pull request :partying_face: 
I merge it. But we need to fix some issues with small pictures on large screens, too. But this should be the next step.

PR #169, opened 2020-06-18
quote reserved words in mysql

tanerkay:
Quote reserved words in MySQL (and potentially mirrored later in mariadb)

Benefits:
- Allows local development using MySQL for those too lazy to spin up an image with MariaDB
- Prevents potential regressions down the track if MariaDB adopts rank as a reserved word (introduced in MySQL 8.0.2)

### DIFF
--- a/src/Entity/Address.php
+++ b/src/Entity/Address.php
@@ -14,7 +14,7 @@ use Doctrine\ORM\Mapping as ORM;
 /**
  * Addresses.
  *
- * @ORM\Table(name="addresses", indexes={@ORM\Index(name="IdMember", columns={"IdMember"}), @ORM\Index(name="IdCity", columns={"IdCity"}), @ORM\Index(name="CityAndRank", columns={"IdCity", "Rank"})})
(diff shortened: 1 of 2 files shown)


tanerkay:
Benefits:
- Allows local development using MySQL for those too lazy to spin up an image with MariaDB
- Prevents regressions down the track if MariaDB adopts `rank` as a reserved word (introduced in MySQL 8.0.2)

thisismeonmounteverest:
@tanerkay I'd expect that Doctrine DBAL takes care of this. But if it doesn't this is fine.
Regarding too lazy to spin up an image with Maria DB: We have a working docker install/setup now.

thisismeonmounteverest:
@vincentchalamon That section of Doctrine's documentation also points out that we could use a QuoteStrategy instead of quoting all columns etc. and hoping that next time we introduce a new table we don't forget about it.

On the other hand, we currently do not have an issue as far as I'm aware, so why bother?

thisismeonmounteverest:
@tanerkay or do you have one when setting up a dev environment because 'rank' isn't quoted?

vincentchalamon:
@thisismeonmounteverest The `QuoteStrategy` would be (IMO) definitively a better solution. @tanerkay If you have a reproducing context, would you give it a try, please, to ensure this strategy is working?

thisismeonmounteverest:
Closing as inactive.

PR #179, opened 2020-07-28
Move about to src

crumbking:
Open points:

- feedback form needs to be moved
- URL FAQ category should start with 0,1,2 ...  instead we have faq/ and faq/2 ...
- Links in about page to BV and BOD pages (external route) work only with login
- Links in translations on second level urls like /about/thepeople are wrong (issue with translation tool?)
- cleanup rest in build/about and legacy loader needed to finish


crumbking:
@thisismeonmounteverest 

Would you mind looking for the remaining first four points?

thisismeonmounteverest:
@crumbking: I fixed the open points 2, 3 and 4 locally. The fourth was quite interesting as the old site set the base URL in the <head> section all relative URLs where relative to that one. The rewrite didn't have that so the urls were relative to /about.

I fix the first one tomorrow.

When will you do the cleanup?

PR #181, opened 2020-08-16
Fix CI

vincentchalamon:
- [x] Fix Coding Style
- [x] Fix Code Sniffer
- [x] Fix Mess Detector
- [x] Fix PHPUnit

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,7 @@
         "nelmio/alice": "^3.0",
         "nikic/php-parser": "^4.0",
         "phploc/phploc": "^6.0",
-        "phpmd/phpmd": "^2.6",
+        "phpmd/phpmd": "^2.6 <2.9.0",
         "phpstan/extension-installer": "*",
         "phpstan/phpstan": "*",
         "phpstan/phpstan-doctrine": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "b8959005e793f77364e28adf773f6319",
+    "content-hash": "34d3bc5d789cdae600a8c38642c72f5e",
     "packages": [
         {
             "name": "amphp/amp",
@@ -10313,16 +10313,16 @@
         },
         {
             "name": "phpmd/phpmd",
-            "version": "2.9.0",
+            "version": "2.8.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpmd/phpmd.git",
-                "reference": "2a346575a45a6f00e631f4d7f3f71b6a05e0d46d"
+                "reference": "714629ed782537f638fe23c4346637659b779a77"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpmd/phpmd/zipball/2a346575a45a6f00e631f4d7f3f71b6a05e0d46d",
-                "reference": "2a346575a45a6f00e631f4d7f3f71b6a05e0d46d",
+                "url": "https://api.github.com/repos/phpmd/phpmd/zipball/714629ed782537f638fe23c4346637659b779a77",
+                "reference": "714629ed782537f638fe23c4346637659b779a77",
                 "shasum": ""
             },
             "require": {
@@ -10333,8 +10333,6 @@
             },
             "require-dev": {
                 "easy-doc/easy-doc": "0.0.0 || ^1.3.2",
-                "ext-json": "*",
-                "ext-simplexml": "*",
                 "gregwar/rst": "^1.0",
                 "mikey179/vfsstream": "^1.6.4",
                 "phpunit/phpunit": "^4.8.36 || ^5.7.27",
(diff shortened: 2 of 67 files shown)


thisismeonmounteverest:
@vincentchalamon #179 has been merged. Please rebase and fix conflicts.

vincentchalamon:
@thisismeonmounteverest Rebased and fixed, it's RFR :-)

thisismeonmounteverest:
@vincentchalamon I get the following error message on `composer install`:

`Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In CheckDefinitionValidityPass.php line 49:
!!

!!    Please add the class to service "app.member_data_extractor.donations" even if it is constructed by a factory since
!!    we might need to add method calls based on compile-time checks.

!!

!!
!!
Script @auto-scripts was called via post-install-cmd`

vincentchalamon:
@thisismeonmounteverest This should be fixed now

thisismeonmounteverest:
@vincentchalamon So I can composer install now and run npm install without any unexpected problems.

But when I try to open localhost:8000 it shows that I do not have a homepage set.

"You're seeing this page because you haven't configured any homepage URL and debug mode is enabled."

vincentchalamon:
@thisismeonmounteverest Indeed, my bad! I've fixed it.
The Symfony 5.1 routing has evolved and I have upgraded the configuration in the Kernel.php file, but I've missed some points. It's fixed now.

thisismeonmounteverest:
@vincentchalamon So logging in works now. So I decided to try https://localhost:8080/mydata which fails with en exception.

`Argument 4 passed to App\Model\MemberModel::__construct() must implement interface Iterator, instance of Symfony\Component\DependencyInjection\Argument\RewindableGenerator given`

vincentchalamon:
@thisismeonmounteverest Does it work now?

Sorry for the ping-pong messages, we really need some functional tests!

thisismeonmounteverest:
@vincentchalamon I agree that we need functional test. But one problem at a time 👍 

I'm currently checking old code against new code (running them against the same database and comparing the results. Slow process.).

vincentchalamon:
@thisismeonmounteverest It's weird, on the original code, the MemberModel::prepareMemberData method has unaccessible code: https://github.com/BeWelcome/rox/blob/bootstrap4/src/Model/MemberModel.php#L602-L610

Could the unaccessible code be removed, or is it a mistake and the first return should be removed?

vincentchalamon:
I've fixed the issues, but I'm stuck with the following error (maybe cause I don't have enough data on local).
Could you try locally with real data?

bewelcome

thisismeonmounteverest:
@vincentchalamon The dead code in the MemberExtractor can be removed.

The Member entity expects that an admin unit can be found for a member. As the seed data doesn't provide any for the created members the code through an exception.

That can be fixed for the test database by running this SQL:

insert into geonamesadminunits set    geonameid =2950157, 
    name=  'Land Berlin',
    fclass= 'A',
    fcode= 'ADM1',
    admin1 = 16,
    moddate = '2020-09-04',
    country = 'DE';
insert into geonamesadminunits set    
    geonameid = 1643012,
    name = 'Provinsi Papua',
    fclass = 'A',
    fcode = 'ADM1',
    country = 'IN',
    admin1 = '36',
    moddate = '2020-09-04';

This solve the issue for the data extractor till we hit the DoctrineFactory for the donations. (Using $this when not in object context.)

In general there are cities with out admin units set in the geonames data dump so a null as admin unit needs to be allowed but that is outside the scope of this pull request.

vincentchalamon:
OK, this is working. I'm gonna push a fix about that.
Locally, I cannot get the zip download because of line https://github.com/BeWelcome/rox/blob/bootstrap4/src/Controller/MemberController.php#L137, I had to remove it. Do you remember why did you set it months ago?

thisismeonmounteverest:
@vincentchalamon I tried to get the page redirect to the profile in case an admin triggered the data export. But that doesn't really work. So I need to change that to the same behavior as the normal data export.

Other than that it works now, but I still want to see the compare so please be patient :)

thisismeonmounteverest:
@vincentchalamon No luck. First I ran into a max_execution timeout which was weird because that was set to 0 in the php.ini so I added set_ini call to set a deliberately high amount. Which work till it tried to extract the newsletters I ever created or translated.

The DoctrineExtractor has two problems: First it has too much donation in it's extraction method, second it assumes that the twig template only use the entity and no other information.

For the Newsletter the member would need to be given as a parameter (named creator). (That's the only one I assume.)

The generated zip files doesn't contain the gallery items.

vincentchalamon:
> First I ran into a max_execution timeout which was weird because that was set to 0 in the php.ini so I added set_ini call to set a deliberately high amount.

The extractor code is the same, it has just been moved to other services. So I guess if you're having a max_execution_time error on extract, you should already have it before this PR.

> The DoctrineExtractor has two problems: First it has too much donation in it's extraction method, second it assumes that the twig template only use the entity and no other information.
For the Newsletter the member would need to be given as a parameter (named creator). (That's the only one I assume.)

I've added a `member` parameter for the template which should fix some of those issues.

What do you mean "_it has too much donation in its extraction method_"? I only see the user donations (in my case: none).

> The generated zip files doesn't contain the gallery items.

I don't see the problem anymore, I've uploaded some pictures on my gallery (member-2) and I can see them in the zip export.

thisismeonmounteverest:
@vincentchalamon As I said the execution timeout shouldn't have happened locally as the execution time was set to 0 which means no timeout. I assume the problem is with the configuration of symfony.

The hint to donations was meant regarding the variable names, which you fixed anyway without understanding the comment.

I merge now because I believe the gallery items problem was due to the wrong data directory.

PR #201, opened 2020-12-07
Fix tests

vincentchalamon:


thisismeonmounteverest:
@vincentchalamon Thanks for your effort to fix those isues.

- How about adding behat tests to the make file?
- I need to check if phpdbg works on Windows (that's the reason for all the spurious php you see)
- The rule for phpcs doesn't match the Github CI setup (--warning-severity=Error should be removed).

vincentchalamon:
> The rule for phpcs doesn't match the Github CI setup (--warning-severity=Error should be removed).

What do you mean? The [GitHub Actions setup](https://github.com/BeWelcome/rox/blob/bootstrap4/.github/workflows/ci.yml#L282) runs the same command than the [Makefile](https://github.com/BeWelcome/rox/blob/bootstrap4/Makefile#L82): `vendor/bin/phpcs --colors --warning-severity=Error`

thisismeonmounteverest:
@vincentchalamon For the last few commits I run make and after pushing either PHP-CS-Fixer failed. The call in the CI run is `vendor/bin/php-cs-fixer fix -v --diff --dry-run`. But might be that the cs-fixer isn't run with make.

vincentchalamon:
@thisismeonmounteverest OK seen! The php-cs-fixer job from Makefile has been fixed.

PR #203, opened 2020-12-07
Fix 'message sent' flash message's translation locale

### LINKED ISSUE #195: Notification in the language of other member
After sending a message, the notification that the message has been sent is written in the default language of the other member, not in my own default language:
Screenshot from 2020-11-25 20-35-25

SzymonKaminski:
This PR aims to close #195.
I explicitly set translation locale for flash in the controller to avoid changing anything else about translator.

### DIFF
--- a/src/Controller/MessageController.php
+++ b/src/Controller/MessageController.php
@@ -184,7 +184,7 @@ class MessageController extends BaseMessageController
             $body = $messageForm->get('message')->getData();
 
             $this->messageModel->addMessage($sender, $receiver, null, $subject, $body);
-            $this->addTranslatedFlash('success', 'flash.message.sent');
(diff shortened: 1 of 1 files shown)


thisismeonmounteverest:
@SzymonKaminski I never thought about that approach. But I think we should fix the root cause and not the symptom. The root cause is a switch of the locale in addMessage (or more likely in the Mailer service). 

There should be no need to change the translator for this. (Though I do not understand why the symfony Mailer doesn't allow to set the locale of the send mails which would solve this problem neatly.)

thisismeonmounteverest:
This got fixed with commit https://github.com/BeWelcome/rox/commit/b9e98858aea9d375c292fea232e309a1c5128543. Closing.

PR #207, opened 2020-12-22
Fix #205

### LINKED ISSUE #205: Remove /api/members/{id} entrypoint
It should not be possible to get a member by its id, for security reasons. Replace the /api/members/{id} by /api/members/{username}

vincentchalamon:


thisismeonmounteverest:
@vincentchalamon If we remove the id route to find members the by country route shouldn't be there either. That makes it way to easy to get the usernames to get details for all members.

Using A-z in a regex matches more than A-Za-z if I'm not mistaken. Maybe use one of these as they represent the real username pattern:
    const PATTERN_USERNAME = '[A-Za-z](?!.*[-_.][-_.])[A-Za-z0-9-._]{2,18}[A-Za-z0-9]';
    const HANDLE_PREGEXP = '/^[A-Za-z](?!.*[-_.][-_.])[A-Za-z0-9-._]{2,18}[A-Za-z0-9]$/';

That would make the tests less brittle at the same time as they would work without the test database as well.

vincentchalamon:
> (...) the by country route shouldn't be there either

@thisismeonmounteverest Do you mean to remove the /api/members route? How will it be possible to look for members then?

> Using A-z in a regex matches more than A-Za-z if I'm not mistaken

Nop, the A-z is a shortcut to A-Za-z. Also, \d means 0-9

thisismeonmounteverest:
@vincent: Ever tried: `preg_match('/[A-z]/', '\\', $matches);` ?

Result of vardump($matches): 
`array(1) {
  [0]=>
  string(1) "\"
}`

Regarding country route and how to find members: That would be the API for search.

The APP would need to be able to show profiles of members only not groups of members therefore at the beginning only the `/api/members/{username}`.

vincentchalamon:
> Ever tried: preg_match('/[A-z]/', '\\', $matches); ?

Interesting... :thinking:

> Regarding country route and how to find members: That would be the API for search.
The APP would need to be able to show profiles of members only not groups of members therefore at the beginning only the /api/members/{username}.

Sorry, I don't understand it. If we want to search for members, it should be /api/members with filters. This entrypoint will return a paginated list of members corresponding to the filters sent.

thisismeonmounteverest:
@vincentchalamon Which route to use when searching is probably a question of taste so I won't discuss that.

Screen scraping is not as simple as API calls. Therefore till we have another concept for authorization I'd just limit the ways to use the API. For now that means no search/filtering options and only the /api/members/{username} route. 

Based on that API the profile pages could be rewritten, and an App that shows your messages and requests could leverage that as well.

vincentchalamon:
> For now that means no search/filtering options and only the /api/members/{username} route.

@thisismeonmounteverest OK I'll remove all routes other than /api/members/{username}. We'll think of a search route later.

> Based on that API the profile pages could be rewritten, and an App that shows your messages and requests could leverage that as well.

So the minimum required API is just the profile pages in order to be used in an app. Sounds like a plan :muscle:

PR #208, opened 2020-12-23
Fix CI

vincentchalamon:


thisismeonmounteverest:
@vincentchalamon Thanks for fixing the CI workflow (seems the xdebug 3 release broke that) and fixing the code duplication in the repositories.

PR #209, opened 2020-12-24
Fix CI: wrong translationAllowed in fixtures

vincentchalamon:


thisismeonmounteverest:
@vincentchalamon I wondered why `make behat` always recreates the database...

PR #233, opened 2021-04-13
Adds react/avatar

pschoffer:
## Description

This is the first step towards handling avatar change in nicer way. With the ability to crop and rotate.

This step introduces React to the mix. Streamlines the profile picture uploading process.

Inspiration for this is the profile picture change here in github.

## Steps to test

Go to edit my profile and click Change my profile picture:
image

## Next steps (seperate PR)
* Adds the cropping rotation popup


src/Controller/AvatarController.php
```diff
@@ -55,6 +92,39 @@ public function showAvatar(string $username, string $size): BinaryFileResponse
         return $this->createCacheableResponse($filename);
     }
 
+    private function storeAvatar($memberId, $tmpFilePath)
+    {
+        // TODO
+        // $this->writeMemberphoto($memberId);
```
pschoffer, line 98: What is the `membersphotos` Photo DB record for? It seems a bit inconsistent here as we add a record but never remove it when deleting old profile picture. Is that by design?

build/members/pages/member.page.php
```diff
@@ -188,6 +188,14 @@ protected function submenu() {
         $member = $this->member;
         $words = $this->getWords();
         $picture_url = 'members/avatar/'.$member->Username;
+        $globalsJs = json_encode([
+            'baseUrl' => $this->getBaseUrl(),
+            'texts' => [
+                'profile.change.avatar' => $words->get('profile.change.avatar'),
+                'profile.change.avatar.success' => $words->get('profile.change.avatar.success'),
```
pschoffer, line 195: How do I define the new translations? `profile.change.avatar.success` `profile.change.avatar.fail`

build/members/pages/member.page.php
```diff
@@ -205,8 +213,7 @@ protected function submenu() {
             if ($this->myself) {
                 // TODO : change language code (en) and wordcode
```
pschoffer, line 214: What does this mean?

build/members/pages/member.page.php
thisismeonmounteverest: That needs to be done manually to update the database to include the translation IDs.

src/Controller/AvatarController.php
thisismeonmounteverest: That is old code. It seems at some point the idea was to upload several profile pictures/images and select one to be used. 

Without an entry in the membersphotos table an empty avatar will be shown.

As you can't decide to not show an avatar/profile picture anymore that entry is never deleted.

build/members/pages/member.page.php
thisismeonmounteverest: The member page is different if you look at it yourself (myself :)) or at someone else's page.

pschoffer: Yeah I got that part, but the `TODO: change language code (en) and wordcode` ?

build/members/pages/member.page.php
pschoffer: Do I do that or somebody does that after merge? :)

src/Controller/AvatarController.php
pschoffer: Oh but looking at the AvatarController:showAvatar that is not used nowadays... so it is probably safe to delete, right?

crumbking:
> Do I do that or somebody does that after merge? :) 

Can be done later on beta. Locally you could use user bwadmin. He has translation rights. In the translation tool in the submenu is the create translation link.

src/Controller/AvatarController.php
thisismeonmounteverest: We shoudl check all usages in queries for membersphotos and then delete it after review.

build/members/pages/member.page.php
thisismeonmounteverest: That's normally done through the translation interface. So I can take that.

build/members/pages/member.page.php
thisismeonmounteverest: Can't say that comment is really, really old.

PR #236, opened 2021-04-30
Cleanup

pschoffer:
There are some CI failures due to code issues.

I plan to go through and address those.

The first one being this duplicate code:
https://github.com/BeWelcome/rox/runs/2402774135

After investigation, It seems, only the code in one or the other class is used as they have parent->child relationship anyway.


src/Controller/BaseHostingRequestAndInvitationController.php
```diff
@@ -132,9 +132,4 @@ protected function getMessageFromData(FormInterface $requestForm, Member $member
 
         return $invitation;
     }
-
-    private function getRedirectUrl(Request $request)
```
pschoffer, line 136: Not used at all (private - must be used within this class, but it is not)

src/Controller/BaseMessageController.php
```diff
@@ -265,84 +265,6 @@ protected function isInvitation(Message $message)
         return null !== $message->getRequest()->getInviteForLeg();
     }
 
-    protected function checkRequestExpired(Message $hostingRequest): bool
```
pschoffer, line 268: All these are called only from `HostingRequestController` and `InvitationController`, both of which inherit directly from `BaseHostingRequestAndInvitationController` (where the duplicate code is). So no need to have it in `BaseMessageController` as well.

pschoffer:
This was now resolved by @thisismeonmounteverest in the original branch.

PR #244, opened 2021-05-19
Avatar save

pschoffer:
This is another step towards having some sort of avatar tweaking tool.

This breaks down avatar change into two steps:
1) Upload the new file
2) Confirm (confirming will move it from temporary space and replace original)

This already brings the value of seeing your new avatar as it will be shown before you confirm the change. 

In the next stage, we could add the ability to rotate/crop avatar on the confirm popup.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -164,6 +164,10 @@ Please read [Useful hints](#useful-hints) section below.
 
 * Production OS is Debian GNU/Linux Strech.
 
+* Logs can be followed by `docker-compose exec php tail -f var/log/dev.log`
+
+* To change log level change [monolog config](./config/packages/dev/monolog.yaml)
+
 ## Create documentation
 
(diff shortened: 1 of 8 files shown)


crumbking:
Hi @pschoffer, I tested it locally. I confirm that I see a popup with the new profile picture. If I directly change the picture once again the popup does not show up and the pic does not change.
But yeah that would be cool to get crop and rotation running....

pschoffer:
Hi @crumbking ,
Thanks for giving it a spin!

What do you mean by:
> If I directly change the picture once again the popup does not show up and the pic does not change.

Is that expected behavior it sounds like a bug right? Could you provide steps to reproduce, please?

crumbking:
Sure,
1. I click on change profile picture and select a new pic
2. I see:
grafik
3. I click save
4. Modal get closed and profile picture does not change even if I get a short "toast" message in green "profile pic updated"

PR #305, opened 2022-10-24
fix: During install, composer is missing from docker image, installing it from correct source …

zeuros:


thisismeonmounteverest:
Hi @zeuros, on which OS are you working? Never had a problem with the docker install and composer.

zeuros:
I tried on windows (which didn't work for few reasons) then ubuntu

Issue #264, opened 2021-11-01
Logo is not readable


hubyhuby:
Just would like to mention the logo is not readable.
1) When on orange background
The orange logo doesn t work. We need a second version of the logo.
![image](https://user-images.githubusercontent.com/26340942/139669609-ab4440c4-424c-4ff8-98e8-3ab118c57771.png)

2) When on white background 
I believe the round world is to complex as a logo. We need something simplier, easier to memorise... like a schematic world..
![image](https://user-images.githubusercontent.com/26340942/139669500-0d43f52a-2143-42bc-9f15-6890802aac97.png)

lajlev:
I did a quick shot on a more minimal logo 
SVG is to be found here https://u.lillefar.dk/fO64EP
<img width="614" alt="image" src="https://user-images.githubusercontent.com/133836/162218292-306ce584-9ad9-4c36-b2ec-1534d98f0884.png">

lajlev:
Took another shot
![image](https://user-images.githubusercontent.com/133836/162375571-feb9bd2c-e339-4492-a3aa-2ccb3e6d82d1.png)

SVG here https://u.lillefar.dk/7Xk7pG

zeuros:
Gave a try adding this logo to app, I don't know how to test mailing yet, so if someone can check this, it could be nice

thisismeonmounteverest:
New logo introduced lately.

PR #306, opened 2022-10-24
#264 feat: changing website logo for a white svg

zeuros:


thisismeonmounteverest:
Hi @zeuros, interesting idea to change the logo to SVG. But shouldn't the logo be replaced with a logo and not a white blob?

thisismeonmounteverest:
@zeuros: Changing the logo through a pull request is not an option. 

Replacing the original one with an svg matching it would be appreciated, though.

zeuros:
When you say the original one, do you mean I should redraw original logo to svg ?

I saw the work made in issue [264](https://github.com/BeWelcome/rox/issues/264) and tried to code it.

If that's not satisfying, just tell what you expect in the issue.

Issue #289, opened 2022-05-20
Use the same Lightbox gallery (with inter-linked images) in the profile's main page than on the gallery page


nekohayo:
If you go to `https://www.bewelcome.org/members/$your_username`, at the bottom there is "Latest pictures". However, they are super tiny (please make them bigger instead of having so much padding around the thumbnails!) and when you click on any of them, you are taken to the individual image's page instead of simply zooming into the image with Lightbox and being able to flip through all the images with the Left/Right arrow keys (or the floating arrows you would get in Lightbox if you tell it to show a collection of images). It's slow and cumbersome (you then have to use the browser's back button to see another image, then click again, etc etc.)

I know you already have the Lightbox code for this because that's exactly how it works in `https://www.bewelcome.org/gallery/show/user/$your_username/images`, so it's a matter of reusing that code directly on the profile page and considering the photos as a linked gallery set (instead of individual images), bringing that UX improvement directly to the profile page, instead of using it only in one hidden place.

thisismeonmounteverest:
This will be implemented when the profile pages code gets rewritten.

zeuros:
Did a PR for that, tell me if you can give it a go

PR #307, opened 2022-10-25
#289 feat: Use the same Lightbox gallery (with inter-linked images) i…

zeuros:
…n the profile's main page than on the gallery page

### DIFF
--- a/build/members/templates/profile.subcolumn_right.php
+++ b/build/members/templates/profile.subcolumn_right.php
@@ -286,21 +286,28 @@ if ($statement) {
             <?php endif; ?>
         </h3>
 
-        <div class="p-2 d-flex flex-wrap justify-content-around">
+        <div class="p-2 d-flex flex-wrap justify-content-around"
+            style="overflow: scroll; max-height: 400px;">
 
-    <?php
-    // if the gallery is NOT empty, go show it
-    $p = PFunctions::paginate($statement, 1, $itemsPerPage = 8);
-    $statement = $p[0];
+            <?php
+            // if the gallery is NOT empty, go show it
+            $galleryImages = PFunctions::paginate($statement, 1, $itemsPerPage = 8)[0];
 
-    foreach ($statement as $d) {
-        echo '<div><a href="gallery/show/image/'.$d->id.'">' .
-           '<img src="gallery/thumbimg?id='.$d->id.'"' .
-               ' alt="image"' .
-               ' style="height: 50px; width: 50px; margin: 1rem;"/>' .
-           '</a></div>';
-    }
-    ?>
+            foreach ($galleryImages as $galleryImage) {?>
+                <a href="gallery/img?id=<?= $galleryImage->id ?>" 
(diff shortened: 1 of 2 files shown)


build/members/templates/profile.subcolumn_right.php
```diff
@@ -286,21 +286,28 @@ class="float-right gray align-self-center"><i class="fa fa-flag" alt="<?=$words-
             <?php endif; ?>
         </h3>
 
-        <div class="p-2 d-flex flex-wrap justify-content-around">
+        <div class="p-2 d-flex flex-wrap justify-content-around"
+            style="overflow: scroll; max-height: 400px;">
```
thisismeonmounteverest, line 290: Isn't a max-height of 400px a bit small?

build/members/templates/profile.subcolumn_right.php
```diff
@@ -286,21 +286,28 @@ class="float-right gray align-self-center"><i class="fa fa-flag" alt="<?=$words-
...
-        echo '<div><a href="gallery/show/image/'.$d->id.'">' .
-           '<img src="gallery/thumbimg?id='.$d->id.'"' .
-               ' alt="image"' .
-               ' style="height: 50px; width: 50px; margin: 1rem;"/>' .
-           '</a></div>';
-    }
-    ?>
+            foreach ($galleryImages as $galleryImage) {?>
+                <a href="gallery/img?id=<?= $galleryImage->id ?>" 
+                    style="max-width: 40%; display: flex; align-items: center; margin-bottom: 1rem;"
```
thisismeonmounteverest, line 298: Why limit the width to 40%? There's likely ample screen space and images are large.

zeuros: This is for x spacing between images / container borders

build/members/templates/profile.subcolumn_right.php
zeuros: Hmmm, I don't know, which height would you suggest ?
Screenshot_20221025_235338

build/members/templates/profile.subcolumn_right.php
thisismeonmounteverest: Obviously it's not a good idea to only look at code. 

If you want to ensure 2 columns using 50% with padding sounds like a good idea.

build/members/templates/profile.subcolumn_right.php
thisismeonmounteverest: As stated below, I think this is fine as it is.

build/members/templates/profile.subcolumn_right.php
thisismeonmounteverest: Let's see what others think about this update...

thisismeonmounteverest:
@zeuros: It's now live on beta. 

Seeing it in use I observed the following: Thumbnails should all have the same height and the lightbox doesn't work as gallery for the up to 8 images but only shows the one clicked on.

Maybe do a second round?

PR #309, opened 2022-11-04
Ability to filter, if persons already host somebody during the searched period

maninthemiddle01:
It would be nice to know before writing a request, if the person, which is requested already hosts someone or is away.
For this I guess it would be necessary to have some kind of internal calendar, where you either put the times you travel or where times, in which a person already hosts someone, are automatically blocked after they confirmed a stay (if this isn't already implemented).
Then it would be nice to have the possibility of a filter, if you look for an accommodation


thisismeonmounteverest:
@maninthemiddle01 This isn't really a pull request is it?

thisismeonmounteverest:
Closing as this is spamming my mailbox.

PR #339, opened 2023-12-09
#248 Fetching activities with at least 0 minutes runtime (instead of only the ones longer than 1 day)

### LINKED ISSUE #248: Improvement to the activity moderation tool
It looks like at the moment, only activities lasting more than one day are shown in the activity moderation tool (this page: https://www.bewelcome.org/admin/spam/activities)
If would be a good idea to show all activities regardless of their duration, since spam activities shorter than 1 day may also be created. Up to now it didn't happen, because I guess spammers prefer to make long-lasting activities to gain visibility. But in case a spammer decides to create a short one in order not to be noticed, we'll look dumb if we're not able to delete it.
Note: the kind of spam activities we typically have to delete are ones created for advertising dating sites.

PLP-GTR:
I have changed the filter for those activities:

> `DATEDIFF(a.ends, a.starts) > 1`

to 

> `TIMESTAMPDIFF(MINUTE, a.starts, a.ends) >= 0`

The seemingly simpler approach to use `DATEDIFF(a.ends, a.starts) >= 0` would have lead to the issue, that activities with an end time prior to the start time on the same day would show up. Although this shouldn't be possible at storage logic already, the "> 1" filter was preventing those from showing up, so I kept this logic.

image

Fixes #248

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -95,5 +95,6 @@ doctrine:
         dql:
             datetime_functions:
                 datediff: App\Doctrine\Functions\DateDiff
+                timestampdiff: App\Doctrine\Functions\TimestampDiff
                 yearweek: App\Doctrine\Functions\YearWeek
                 geodistance: App\Doctrine\Functions\GeoDistance
--- a/src/Doctrine/Functions/TimestampDiff.php
+++ b/src/Doctrine/Functions/TimestampDiff.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace App\Doctrine\Functions;
+
+use Doctrine\ORM\Query\AST\Functions\FunctionNode;
+use Doctrine\ORM\Query\AST\Node;
+use Doctrine\ORM\Query\Lexer;
+use Doctrine\ORM\Query\Parser;
+use Doctrine\ORM\Query\SqlWalker;
+
+/**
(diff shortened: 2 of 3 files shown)


thisismeonmounteverest:
@PLP-GTR Why not just remove the datediff? The original idea was to find all activities that have been created by a member that has been banned because of spamming and that lasted longer than a day. Issue #248 says just show all activities of banned members to remove them if necessary. No need to check anything else.

thisismeonmounteverest:
@PLP-GTR Regarding the heads up: In the dev environment most of the times newly introduced classes are found automatically. In the few cases that doesn't help you need to call php bin/console cache:clear. (Which is also needed to create the cache on production.)

PLP-GTR:
> @PLP-GTR Why not just remove the datediff? The original idea was to find all activities that have been created by a member that has been banned because of spamming and that lasted longer than a day. Issue #248 says just show all activities of banned members to remove them if necessary. No need to check anything else.

Three reasons:
- I did not know the reason of the initial intention to filter only activities with a date (day) difference of at least `+1`
- I did not want to include the negative activities since I don't know what the current production database holds (worst case thousands of activities with negative time)
- I think it's a sexy solution to include the TIMESTAMPDIFF method and let the initial intention live on (plus it was fun to learn about doctrine/DQL)

I can change the handling if you want, no problem. Let me update #248 and draft this pull request.

PR #357, opened 2024-08-27
Added anti-spam checks

gabriel-BW:
Added check for phone numbers in comments, and for phone numbers and email addresses in relation requests

### DIFF
--- a/src/Controller/CommentController.php
+++ b/src/Controller/CommentController.php
@@ -171,6 +171,7 @@ class CommentController extends AbstractController
             if (
                 $commentModel->checkCommentSpam($loggedInMember, $comment)
                 || $commentModel->checkForEmailAddress($comment)
+                || $commentModel->checkForPhoneNumber($comment)
             ) {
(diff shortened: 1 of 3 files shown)


thisismeonmounteverest (commented):
For which country are the telephone numbers suppose to match?

I would also assume that most people use a mobile phone scheme like +49 12 34 56 78.

gabriel-BW:
That regexp will detect a series of 8 or more digits, with or without intermediate signs (spaces, dots, hyphens and closing bracket which is often included after an optional part at the beginning of a phone number). I disregarded the original + with international code and opening bracket, considering that at least 8 digits should follow anyway. And people from the same country contacting each other wouldn't include the +xx.
That should cover the syntax of phone numbers from most countries, including the standard example you gave. And I don't see other use cases to insert a series of 8 digits in a comment or friend request.
If you have doubts, you can check the DB for comments and friend requests matching this regexp and see if there are any results other than phone numbers.

PR #359, opened 2025-04-13
Small changes (spam filters & profile preferences)

gabriel-BW:
Updated the spam filter for phone numbers (which could match a date), and changed the default profile preference to not accepting contact from empty profiles


leon-wbr (approved):

thisismeonmounteverest:
I incorporated the changes into the codebase. Therefore closing here as code is too far away from develop now.

Issue #360, opened 2025-04-23
Ability to use TOTP 2FA multi-factor authentication


nekohayo:
Hello there, this is a security/safety feature suggestion: as the modern world is not as safe as it used to be, it would be desirable for BW's website to offer, the opt-in ability for users to require successful 2FA using the [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password) 2FA standard, on each login.

That technique/technology is generic and lets you use any independent authentication application (see the [list of such applications](https://en.wikipedia.org/wiki/Comparison_of_OTP_applications)).  
For example, [Aegis](https://getaegis.app) on an Android phone or GNOME's [Authenticator](https://apps.gnome.org/Authenticator/) on a laptop (or Linux phone), etc.

The idea is simple: during your travels, if someone steals (or borrows) the laptop you use to log into the web interface, they would not be able to log into the BW website without also having stolen and unlocked your second device that has the 2FA, or vice versa. Or they might have to go through more steps, at least.

As BW is kind of like a social network or online identity, it might be desirable to be able to protect it that way, for those who feel confident about using this type of added security.

PR #365, opened 2025-08-17
feat: Initial 2FA implementation with Scheb\TwoFactorBundle

jj2bw:
Goal: support 2FA/MFA as any website containing personal information should, even if no financial/health data.

Implementation
Following https://symfony.com/bundles/SchebTwoFactorBundle/7.x/installation.html
Reviewed https://symfony.com/bundles/SchebTwoFactorBundle/current/troubleshooting.html

Work In Progress
Limited testing as issues with local docker and CI (separate PR/issues)
Current CI results with ci changes: https://github.com/jj2bw/rox/actions/runs/17011983649

Would address https://github.com/BeWelcome/rox/issues/360


config/packages/scheb_2fa.yaml
```diff
@@ -0,0 +1,23 @@
+# config/packages/scheb_2fa.yaml
+scheb_two_factor:
+    backup_codes:
+        enabled: true
+    totp:
+        enabled: true                  # If TOTP authentication should be enabled, default false
+        server_name: bewelcome.com     # Server name used in QR code
```
thisismeonmounteverest: Nope. The server would be bewelcome.org. But I suppose it is better to use a parameter here.

config/packages/scheb_2fa.yaml
```diff
@@ -0,0 +1,23 @@
...
+        issuer: BeWelcome              # Issuer name used in QR code
+        leeway: 60                     # Acceptable time drift in seconds, must be less or equal than the TOTP period
+        parameters:                    # Additional parameters added in the QR code
+            # image: 'https://my-service/img/logo.png'
+            image: 'https://miro.medium.com/v2/resize:fill:128:128/1*65AfOY_oNSTe2G1bFMwQ4A.jpeg'
```
thisismeonmounteverest, line 12: This should be served through our own server. Could this be done relative to the base url?

config/packages/scheb_2fa.yaml
```diff
@@ -0,0 +1,23 @@
...
+        template: security/2fa_form.html.twig
+    google:
```
thisismeonmounteverest, line 14: Didn't you say you wanted to try with github.com first?

src/Entity/Member.php
thisismeonmounteverest, line 1: This introduced new columns for the member entity but there is no migration. Additionally the two factor auth should be optional which means that there needs to be a preference to enable it and the columns should be in a table of their own (we already have way too many columns on member).

templates/security/2fa_form.html.twig
```diff
@@ -0,0 +1,39 @@
+{% extends "base.html.twig" %}
```
thisismeonmounteverest, line 1: This is lacking translations.

thisismeonmounteverest (changes requested):
Thanks for working on this. But needs more work to be able to be pulled.

config/packages/scheb_2fa.yaml
jj2bw: for now switched to .org
not sure where variable should be defined.

config/packages/scheb_2fa.yaml
jj2bw: just to know which file. I got this one while doing image search for a small logo

config/packages/scheb_2fa.yaml
jj2bw: Github first is for the sso.
No github authenticator app option AFAIK ;)

src/Entity/Member.php
jj2bw: imho, that should come after validating things are functional. more database optimisation.

templates/security/2fa_form.html.twig
jj2bw: similary, after validating things are functional (and not breaking anything) as this initial work and there will certainly be polishing later. But no point to polish something not working.

jj2bw:
> Thanks for working on this. But needs more work to be able to be pulled.

This is expected as Work In Progress.
Thanks for the feedback

thisismeonmounteverest:
No activity since 7 months. Closed.

PR #367, opened 2025-08-17
ci(ci): update and pin github actions, use env, permissions

jj2bw:
Current CI is disabled with no run in history (https://github.com/BeWelcome/rox/actions/workflows/ci.yml "This scheduled workflow is disabled because scheduled workflows are disabled by default in forks." but not a fork?) and using old/deprecated action versions that would not run.
This makes CI functional again.

Resources
https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide
If enabled, Github code scanning recommends pinning actions and enforcing workflows permissions which PR does.


jj2bw:
Thanks for the merges!

PR #398, opened 2026-04-08
UX/UI redesign of the header, frontpage & footer

ssolida:
We are making the site easier to read and use. The top bar was too busy, so we moved extra links to the footer. This keeps the layout calm and simple for a community project. The next step will be to improve search in the top menu.

We also redesigned the footer. It has better contrast and is easier to read. The content is split into clear sections, so the page looks tidier and simpler to scan. Language choice and legal links are in a sensible place.

Please try the site as a guest and as a logged in member, and check the footer on a small screen if you can.

Thank you for reviewing this.
Screenshot 2026-04-09 at 01 41 41
Screenshot 2026-04-09 at 01 42 46


dmcfaddengalway:
Can I ask why are there 3 footer files?

assets/scss/_footer.scss
```diff
@@ -6,37 +6,355 @@ footer {
...
+      gap: 0.45rem;
+      padding: 0.4rem 0.75rem;
+      font-size: 0.8125rem;
+      font-weight: 500;
+      line-height: 1.35;
+      color: #c9c9ce;
+      text-decoration: none;
+      cursor: pointer;
+      background: rgba(255, 255, 255, 0.05);
+      border: 1px solid rgba(255, 255, 255, 0.12);
```
dmcfaddengalway, line 292: It might be good for us to also setup css var files to establish common colours, border radiuses, etc...

ssolida: Footer styles are footer-only and not shared elsewhere, so CSS variables aren’t a must for this PR. you're right and If we do a wider design pass on common UI later, then centralising colours and radiuses etc in variables would be worth it I guess

thisismeonmounteverest:
private/footer.html.twig should stay as is for now. That is only used when you export your data. So the footer has a different purpose here.

thisismeonmounteverest:
I like the clean structure of the footer and how it changes when the page gets smaller. It definitely looks cleaner that way.

But I also think that people won't use any of the links that are now in the footer because they are too far away from the top.

Additionally, avoid names like footer-modern-heading in the near future something else will be modern and the 'modern' just adds noise reducing readability.

The select to activate translation mode looks a bit skewed (login as bwadmin to see that).

Something weird is going on with the symfony toolbar as it can't be collapsed anymore now.

ssolida:
@thisismeonmounteverest Thanks for the detailed review, it really helps.

I’ve left private/footer.html.twig as it was, since it’s only for the data export flow and that footer has a different job there.

I’m glad the cleaner structure and how it behaves on smaller screens work for you.

I’ve also wired the same shared footer into the logged-out homepage, so it matches the rest of the site now. Feels consistent end to end.

On the point about links being far from the top: I agree that distance from the header affects how much people notice them. What we’re doing here is still a step forward for accessibility, navigation, and having a clear footer, which matters. For things that feel more secondary, we could surface them closer to the fold later (homepage highlights, a blog block, or something like that). For now I think this footer is the right baseline: one consistent, readable layout instead of spreading links around without a clear hierarchy.

I’ve adjusted the translation mode select so it sits straight for bwadmin; if anything still looks off, I’m happy to tweak it.

On the Symfony toolbar not collapsing: I’ve seen the same thing on my side, and it doesn’t look related to this PR from what I can tell.

Thanks again.

ssolida:
After this PR is merged, I’d like to work on the homepage next.

Right now it feels old: some pictures look dated, blurry, or pixelated on normal screens. Parallax was a trend a few years ago, but today I think it does not help much: it makes the page harder to read, it is not great for accessibility, and it usually adds more JS/CSS that is harder to keep simple and clean.

I’d prefer a clearer, simpler layout. The homepage is the first thing people see when they open the site, so it should look modern and easy to use. For the visuals, I’m also thinking about illustrations or simple graphics instead of old photos that do not look good anymore.

I already have ideas for a follow-up PR. This PR stays focused on the footer and the fixes we discussed here.

thisismeonmounteverest:
@ssolida The Symfony toolbar collapses just fine without the footer changes, so I do think it is related.

The new footer on the home page (when not logged in) doesn't work for me at all. Clicking any link there will ask to login and non members will likely not like that.

Translation mode is still askew:

image

How do you think that people will use the forum pages if they only find the link to them at the bottom of the page? I think that is even worse if you opened a forum thread with loads of posts and what to switch to something else, you're going to scroll a lot. We might need a button to jump to the footer :)

ssolida:
Thanks @thisismeonmounteverest  for taking the time to review. 
Same here on the Symfony toolbar: in dev it misbehaves even without my branch, so I am not digging into that for this PR and I do not see it as a blocker :)

The guest home footer now shows the same public links as the current footer (safety, get answers, legal, language...), so behaviour stays aligned with what we already ship.

Translation mode was reworked to match the help translate chip and looks cleaner now

On forums and the footer, this change is only a first step and stays within footer scope. 
Logged in members already reach forums from the dashboard, so the footer is not meant to be the only entry point. A broader redesign of the connected homepage would still make sense later, for example quick access to key areas, and I would be glad to review that from a usability angle when it is tackled.

Thanks again.

bohwaz:
That footer is great but I don't think that removing many useful links from the header is a good idea :-/

The header definitely needs some UX work, but removing useful navigation that needs to be there is not the way to proceed IMHO.

ssolida:
> That footer is great but I don't think that removing many useful links from the header is a good idea :-/
> 
> The header definitely needs some UX work, but removing useful navigation that needs to be there is not the way to proceed IMHO.

Thanks for the feedback. I agree the top navigation will need another pass later.

The goal here is intentionally a lighter header. Before, it was crowded: logo plus five separate menus, plus messages, plus profile. It was hard to read and use.

Things like forums, wiki, about, and the rest do not need to sit in the header for this step. They are already on the homepage and in the footer on purpose. The slim header stays for what matters first: brand, message, profile. I will revisit the top menu in a follow up. For now this is the tradeoff that keeps the bar simple and the rest of the journey in the page and footer. We can add something like a quick access block on the home later if we want one clear hub without turning the header back into a mega menu.

bohwaz:
I disagree, specifically because the homepage navigation is confusing.

Removing the navigation in the header doesn't bring anything useful. Having 5 items in the header is not too much. And the footer is not a useful navigation, no one ever goes there to find things (except for contact or legal page).

Things should be done the other way around:
- revamp the homepage (currently it's messy and confusing)
- think and discuss of what should be in the header navigation
- revamp the navigation

This is too early, and this change would make navigation in the site more complex. Removing navigation is not good UX. Changing it to something better is :)

ssolida:
Hi @bohwaz,

Thanks a lot for taking the time to reply, I really appreciate it.

I get your concern about navigation and the homepage. One clarification so we’re not talking past each other: what felt heavy to me wasn’t “five links”, it’s five top-level menus, each with its own subtree and dropdown. That’s a lot of visual and cognitive load on a first visit, and that’s what I meant when I said the bar felt crowded.

On how I’m approaching this: I’m not trying to ship “the whole new site” in a single PR. I prefer small, incremental steps so each change stays easy to review and we can actually move forward. This PR is one of those steps (clearer paths through the page and footer). Nothing here removes access to content in a meaningful way: the links already exist elsewhere in the flow we’re setting up, and this is shippable from that perspective.

For context, I’ve already started reworking the logged-out front page because it felt cluttered and a bit dated to me, and I’m aiming for something simpler and more modern, with a lighter header for this phase. I’ve been working as a frontend developer and UX/UI designer for about twenty years; for now I stand behind this direction, and of course if someone wants to propose something better later, that can always be iterated on (fork, follow-up PR, etc.). What I’m trying to avoid is endless discussion with no merge, especially since I’m contributing on my free time and I want my help to land concretely.

The logged-in homepage will need more time and will come in a follow-up; that doesn’t block shipping this. I’m adding a screenshot of the current logged-out homepage (work in progress) so you can see where things are heading.

I’d really like to merge and close this PR so we can keep momentum. I’m happy to share more before/after screenshots as the homepage work progresses.

Thanks again for the feedback.

Screenshot 2026-04-12 at 21 52 50
Screenshot 2026-04-12 at 22 10 32

bohwaz:
I agree that the current menu is bad, but no navigation to important features is worse.

I opened an issue to discuss the header navigation separately: https://github.com/BeWelcome/rox/issues/399

I propose that you split the header change in a separate PR, so that you can get the footer done quickly :)

That logged-out screenshot looks great btw! Thanks for your work!

thisismeonmounteverest:
I agree with bohwaz: We should make this a package not a box of pieces. 

Changing the header without replacing that navigation or just moving it to the bottom of the page making it hard to accessible to the user might look nicer but doesn't increase the UX.

Regarding the translation mode, that looks out of place with the different color now. The language icon doesn't add anything, does it?

image

As for my comments about naming: Those are serious. Please rename the classes.

assets/scss/_footer.scss
dmcfaddengalway: Makes sense! I agree we can come back to that later.

dmcfaddengalway:
I think this is really great work @ssolida! Thank you! 
Most of the links in the header were not frequently visited links and the footer is a much better place for them.

"Having 5 items in the header is not too much. And the footer is not a useful navigation, no one ever goes there to find things (except for contact or legal page)."

5 items in my opinion is definitely TOO much.  The footer is the perfect place then for contact, safety, legal etc. For sure I can state that maybe Community moves back up to the header bar, but for now I think the cleanliness is nice.

I agree with @ssolida that in this step a lighter header bar is a step in the right direction.

dmcfaddengalway (approved):

ssolida:
Following what we agreed and the discussions around Community links (and the concern that tightening the bar could hide or dilute them), here is the approach I’m going with now, and I think it’s the better one.

Find & Community are icon actions next to messages. Nothing is removed: forums, groups, activities, and wiki stay one tap away from the same Community entry. Only the presentation changes (icons + sheet on small screens), not the destinations.

Mobile follows common app patterns: full-height sheets for Search, Community, and Profile (below xl), with a close control and readable spacing.

Messages got a lighter refresh: a clearer badge when you have unread mail.

Logo: the “BeWelcome” wordmark can stay visible on smaller screens; I tuned padding and margins so the bar stays legible and nothing sits flush against the screen edge.

So the bar stays clear, adaptive, and honest about Community: same links, clearer chrome.

Next: I’ll do the logged-out / front-page navigation and a refresh of the home page in the next steps, once we agree on that one.

These commits include everything above. I also removed the old hamburger flow: it wasn’t really usable on mobile anymore.

here are the screenshots of actual vs this proposal.

ssolida:
> I agree with bohwaz: We should make this a package not a box of pieces.
> 
> Changing the header without replacing that navigation or just moving it to the bottom of the page making it hard to accessible to the user might look nicer but doesn't increase the UX.
> 
> Regarding the translation mode, that looks out of place with the different color now. The language icon doesn't add anything, does it?
> 
> image
> As for my comments about naming: Those are serious. Please rename the classes.

I've fixed that as well

thisismeonmounteverest:
This really looks great. The separators we have at the current site do not make sense anymore with the larger selection area for the menu items I think (and they are too light to see anyway :)). 

Wiki can go to the footer. That is seldom used anyway.

For the menus on mobile I get an error message. But on the thumbnails it feels like the header is too light for the dark background.

PR #402, opened 2026-04-15
Replaced yarn with bun, fixed missing flags and updated dependencies

elcreator:


thisismeonmounteverest:
@elcreator The flags are not missing. They are generated from circle-flags which uses symbolic links for some flags as they are the same. That fails on Windows and it seems MacOS. In the docker container that shouldn't be a problem.

PR #410, opened 2026-04-25
Fix translation duplication 

elcreator:
which makes i.e. key "Armenian" untranslated in Ukrainian version due to duplication in DB

### DIFF
--- a/src/Command/ManticoreIndicesForumCommand.php
+++ b/src/Command/ManticoreIndicesForumCommand.php
@@ -64,6 +64,7 @@ class ManticoreIndicesForumCommand extends Command
         $index = $client->table('forum_rt');
 
         try {
+            $index->drop(true);
             $index->create(
                 [
                     'post_id' => ['type' => 'integer'],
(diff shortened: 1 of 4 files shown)


thisismeonmounteverest:
This pull requests states to do something about translations but changes the index creation for Manticore.

Additionally it is unclear how sorting the found translations solves the issue. The missing locale will be dealt with in the changes necessary to reflect the changes in the database structure.

elcreator:
@thisismeonmounteverest https://github.com/BeWelcome/rox/pull/410/changes#diff-eed8775b414705c50e71e8355d849d6161df5b0deafa4c15b913510effa65d65R55 this line adds predictable sorting.
Unfortunately there is always one translation key which can't go because all fixes are applied to newer version
_zC4nLfLk3dByT5Q

https://github.com/BeWelcome/rox/pull/410/changes#diff-2b00e56ac15df3357a04d699b65b5bcc4cb05188e9d5cb52851ba33162066a7fR41 this line is bugfix for https://github.com/BeWelcome/rox/pull/410/changes#diff-2b00e56ac15df3357a04d699b65b5bcc4cb05188e9d5cb52851ba33162066a7fR33 missing parameter which also affects translation logic by wrong count due to NULL instead of actual param. 

Two additional ones about index dropping are not related, they just gave warnings I noticed during this bugfix.

thisismeonmounteverest:
@elcreator As said, unrelated things should be in an unrelated pull request. When I wrote the commands there was no way to check if th eindex existed. Your approach to just drop the index looks promising if that always works even if the index isn't there yet.

Translation: If there are duplicates than that is the real issue not sorting or the missing parameter. I will take care of the missing parameter when I deal with the database changes. I might even add sorting but I still do not see why that would solve the Armenian translation issue.

PR #413, opened 2026-05-08
new signup and finalize page

ssolida:
Redesigned and modernized the signup and finalize pages UI


translations/validators.en.yaml
thisismeonmounteverest, line 1: This file will have no impact.

assets/js/accommodation_widget.js
```diff
@@ -39,15 +41,21 @@ const slider = document.querySelectorAll('input[type="range"]');
...
     const valueOutput = document.getElementsByClassName('rangeSlider__value-output');
+    const markerIndex = Math.max(0, Math.min(markers.length - 1, Number(value)));
     if (valueOutput.length) {
-        valueOutput[0].innerHTML = markers[value];
+        valueOutput[0].innerHTML = markers[markerIndex];
     }
 }
 
 const initializeHostingInterestSlider = () => {
+    if (slider.length && Number.parseInt(slider[0].value ?? '0', 10) <= 0) {
```
thisismeonmounteverest, line 51: The slider intentionally should not start with 5 but force the member to select a value. Otherwise we end up with everyone having a 5.

PR #429, opened 2026-05-28
Defer Manticore indexing from app entrypoint

### LINKED ISSUE #428: docker-entrypoint should not crash app when manticore startup indexing fails
## Summary

The `app` container entrypoint (`docker/php/docker-entrypoint.sh`) used to run `bin/console manticore:indices:forum` and `bin/console manticore:indices:geonames` before starting FrankenPHP. When MariaDB was reachable but legacy schema was incomplete (e.g. during stage mock DB seeding or production recovery while a Hetzner volume is still loading), indexing failed and—combined with `set -e`—aborted the entrypoint.

**Observed error:** `SQLSTATE[42S02]: Base table or view not found` for `forums_posts` during `manticore:indices:forum`.

**Observed symptom:** Container exits with code **255** even though Symfony could otherwise boot (DB wait, migrations, cache warmup succeeded or were not the failing step).

Search indexing and HTTP serving are different readiness concerns; coupling them caused restart loops while the database or search layer was not yet ready.

## Impact

- **Stage:** Mock or partial DB can be up before forum/geonames tables exist; app never reaches FrankenPHP.
- **Production (steady):** Schema mismatch fails the web container instead of a dedicated index job.
- **Production (recovery):** After host/DB restart, MariaDB may accept `SELECT 1` before datadir on a Hetzner volume is fully usable; indexing or migrations against a half-ready DB worsen the failure mode.
- **Operations:** Unclear logs; no separate step to build indices after DB is actually ready.

## Reproduction

1. Deploy or restart the stack with `APP_ENV=prod` (or dev with startup indexing enabled).
2. Use a database that is reachable but missing `forums_posts` (partial import, empty instance, or volume still loading).
3. Start the `app` service with an entrypoint that runs Manticore index commands before `exec frankenphp`.

**Expected logs (historical):** failure in `manticore:indices:forum` → entrypoint exit → container restart loop.

## Expected behavior

| Concern | Behavior |
|--------|----------|
| **HTTP / FrankenPHP** | Start after DB connectivity (and migrations where appropriate). **Do not** exit solely because Manticore indexing failed or was skipped. |
| **Manticore / search** | May be unavailable until indices exist; failures must be **logged clearly**, not fatal to the web container. |
| **Stage** | App up with mock DB; indexer can wait briefly for sentinel tables. |
| **Prod (steady)** | DB on volume assumed ready at deploy; indexer fail-fast if schema is wrong. |
| **Prod (recovery)** | Optional wait for schema while volume loads (`BEWELCOME_DB_RECOVERY_MODE`); indexer as a separate job after DB is ready. |
| **Observability** | Logs state skip vs run vs failure; alerts on failed index job (follow-up, not blocking this issue). |

## Proposed solution

1. **App entrypoint (prod):** Default `SKIP_MANTICORE_ON_STARTUP=1` when `APP_ENV=prod`. If indexing runs (dev), failures are non-fatal with explicit log lines.
2. **One-shot `manticore-index` Compose service:** `docker compose run --rm manticore-index` — table-wait policy via `BEWELCOME_DEPLOY_ENV` / `BEWELCOME_DB_RECOVERY_MODE`; exits non-zero on real failure so deploy can fail without killing `app`.
3. **`BEWELCOME_DB_RECOVERY_MODE`:** When `1`, app and indexer poll for `BEWELCOME_DB_READY_TABLES` (default `forums_posts geo__names`) before migrations/indexing (Hetzner volume recovery).
4. **Docs:** `doc/docker-deployment.md` — steady prod, stage, recovery playbook, reindex commands.

**Reindex after DB is ready:**

```bash
docker compose run --rm manticore-index
# break-glass:
docker compose exec app bin/console manticore:indices:forum
docker compose exec app bin/console manticore:indices:geonames
```

## Acceptance criteria

- [ ] `app` container does **not** exit solely due to Manticore indexing failure when `APP_ENV=prod` (indexing skipped on boot).
- [ ] When indexing is attempted (dev / explicit override), failures are logged and FrankenPHP still starts.
- [ ] Logs clearly state **SKIPPED**, **running**, **succeeded**, or **failed** for startup indexing.
- [ ] `manticore-index` job builds forum + geonames indices; exits 0 on success, non-zero on timeout/command failure.
- [ ] `doc/docker-deployment.md` documents deploy modes and reindex commands.
- [ ] `./docker/scripts/verify-manticore-startup.sh` documents smoke checks.

## Related context

- Manticore indices are built **from** MariaDB via console commands; MariaDB does not wait on Manticore.
- Compose `depends_on: manticore` is container start order only.
- Alpha stage/prod use `APP_ENV=prod`; use `BEWELCOME_DEPLOY_ENV` and `BEWELCOME_DB_RECOVERY_MODE` for environment-specific behaviour.
- Follow-up: deploy automation + alerting when `manticore-index` fails (sysadmins-infra / observability).

Neophytis:
Skip startup indexing when APP_ENV=prod so FrankenPHP is not killed by missing forum tables during partial DB or volume recovery. Add a manticore-index one-shot Compose job, shared DB wait helpers, and BEWELCOME_DB_RECOVERY_MODE for Hetzner volume restarts.

Fixes #428

### DIFF
--- a/docker-compose.override.yml.dist
+++ b/docker-compose.override.yml.dist
@@ -1,9 +1,10 @@
+# Copy to docker-compose.override.yml for local development (see Makefile install).
 services:
   app:
-    ports:
-      - target: 80
-        published: 8080
-        protocol: tcp
+    build:
+      target: bewelcome_php_dev
+    volumes:
+      - .:/srv/bewelcome:rw,cached
 
   mailer:
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
   app:
     build:
       context: .
-      target: bewelcome_php_dev
+      target: bewelcome_php
       <<: *cache-from
     image: ${APP_IMAGE:-bewelcome/app}
     environment:
@@ -22,10 +22,8 @@ services:
       - db
       - manticore
       - mailer
-    volumes:
-      - .:/srv/bewelcome:rw,cached
-      # if you develop on Linux, you may use a bind-mounted host directory instead
-      # - ./var:/srv/bewelcome/var:rw
+    # Production image runs from baked-in /srv/bewelcome (vendor included).
+    # Local development bind-mounts the repo via docker-compose.override.yml.
     ports:
       - "8080:80"
       - "8043:443"
--- a/docker/php/docker-entrypoint.sh
+++ b/docker/php/docker-entrypoint.sh
@@ -6,6 +6,19 @@ if [ "${1#-}" != "$1" ]; then
 	set -- frankenphp run "$@"
 fi
 
+ensure_composer_dependencies() {
+	if [ -f vendor/autoload_runtime.php ]; then
+		return 0
+	fi
+
+	echo "Installing Composer dependencies (vendor/autoload_runtime.php missing)..."
+	if [ "$APP_ENV" = 'prod' ]; then
+		composer install --prefer-dist --no-dev --no-progress --no-interaction --no-scripts
+	else
+		composer install --prefer-dist --no-progress --no-interaction --no-scripts
+	fi
+}
+
 if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ] || [ "$1" = 'ls' ]; then
  	PHP_INI_RECOMMENDED="$PHP_INI_DIR/php.ini-production"
 	if [ "$APP_ENV" != 'prod' ]; then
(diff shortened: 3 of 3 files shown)


thisismeonmounteverest:
Changing composer.json shouldn't be necessary (instead running composer update should write the correct versions into composer.lock; done already on develop, so no longer necessary). 

Same for package.json only forced overrides are necessary but not changing version info (webpack-dev-server).

Neophytis:
> Changing composer.json shouldn't be necessary (instead running composer update should write the correct versions into composer.lock; done already on develop, so no longer necessary).
> 
> Same for package.json only forced overrides are necessary but not changing version info (webpack-dev-server).

For the Docker/production work, we can drop composer.json / composer.lock if develop already has the security updates. For JS, we only need the overrides block (and bun.lock refresh), not the webpack-dev-server version bump.

PR #431, opened 2026-05-29
Build frontend assets into production image (follow-up to #429)

Neophytis:
## Summary

Follow-up to #429. The production `bewelcome_php` image was missing `public/build/entrypoints.json`, so Symfony threw a Webpack Encore entrypoints error at runtime.

- Enable the `bun encore production` build in the `bewelcome_php` stage so `public/build/entrypoints.json` and hashed assets are baked into the image.
- Add a tiny `var/translations/index.js` stub before the Encore build so webpack can compile at build time (no DB needed). The entrypoint's runtime cache warmup populates real translations from the database.
- Document the production vs. dev asset workflow in `docker/DEPLOYMENT.md` (linked from `INSTALL.md`).

## Test plan

- [x] `docker build --target bewelcome_php -t bewelcome/app .` succeeds
- [x] `docker run --rm --entrypoint sh bewelcome/app -c 'test -f public/build/entrypoints.json'` passes
- [x] Deployed to alpha.stage: homepage returns HTTP/2 200 and serves hashed `/build/*` assets (no Webpack entrypoints error)
- [x] Dev workflow unchanged (override builds `bewelcome_php_dev`, entrypoint runs `bun encore dev`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,15 +94,25 @@ RUN set -eux; \
 	composer install --prefer-dist --no-dev --no-scripts --no-progress --no-suggest; \
 	composer clear-cache
 
-# prevent the reinstallation of node_modules at every changes in the source code
-#COPY package.json yarn.lock webpack.config.js postcss.config.js tailwind.config.js tsconfig.json ./
-#RUN set -eux; \
-#	bun i --frozen-lockfile; \
-#	bun encore production --mode=production
+# Stub var/translations so encore can compile (entrypoint cache warmup fills real translations at runtime)
+RUN mkdir -p var/translations && printf '%s\n' \
+	'// auto-generated stub for image build' \
+	'export const localeFallbacks = {"en":"en"};' \
+	'export const messages = {};' \
+	> var/translations/index.js
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -94,15 +94,24 @@ RUN set -eux; \
...
 # prevent the reinstallation of node_modules at every changes in the source code
-#COPY package.json yarn.lock webpack.config.js postcss.config.js tailwind.config.js tsconfig.json ./
-#RUN set -eux; \
-#	bun i --frozen-lockfile; \
-#	bun encore production --mode=production
+COPY package.json bun.lock webpack.config.js postcss.config.js tailwind.config.js tsconfig.json ./
+RUN set -eux; \
+	bun i --frozen-lockfile; \
+	bun encore production --mode=production; \
+	rm -rf node_modules
```
thisismeonmounteverest, line 110: How does deleting node_modules prevent reinstallation?

Neophytis: it doesn't this is confusing. it's layer caching and removing the folder so the final image size is reduced. For Production you only need the static files usually not the setup files.

PR #438, opened 2026-06-21
Fix Trivy Alpine CVEs and CI wait deadlock

Neophytis:
## TL;DR

- **Trivy:** run `apk upgrade` in the Dockerfile so Alpine security patches (curl, libcap, musl, nghttp2) are applied before the image is scanned
- **CI wait:** stop using `wait-on-check-action` (it waited on *all* checks and could run forever); deploy now polls the `ci.yml` workflow on the same commit, only after the image scan passes

## Why CI wait was stuck

`lewagon/wait-on-check-action` waits until **every check** on the commit passes, except checks from the current workflow. In practice it kept polling long after **CI** had finished — likely because other checks from this workflow or unrelated runs were still pending in GitHub's Checks API.

The fix:
1. Remove the standalone **Wait for CI** job (no more parallel polling for 10+ minutes)
2. In **Deploy to stage**, poll `ci.yml` by commit SHA — CI is almost always done by then anyway (build + scan take longer)

## Trivy findings addressed

All reported CVEs were in Alpine 3.23 base packages with fixes already in the repos:

| Package | Severity | Fixed in |
|---------|----------|----------|
| curl / libcurl | MEDIUM | 8.19.0-r0 |
| libcap | HIGH | 2.78-r0 |
| musl-utils | HIGH/MEDIUM | 1.2.5-r22/r23 |
| nghttp2-libs | HIGH | 1.68.1 |

`apk upgrade --no-cache` at the end of the production stage pulls these in.

## Test plan

- [ ] **Build and publish image** completes **Scan image** without CVE failures
- [ ] **Deploy to stage** CI wait step exits quickly when CI already passed
- [ ] Stage deploy still blocked if CI failed on the same commit

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -69,10 +69,10 @@ jobs:
             runner: ubuntu-24.04-arm
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # tag=v4.2.2
 
       - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772  # tag=v3.4.0
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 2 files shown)


thisismeonmounteverest:
Never use versions tags on actions. Tags can be easily changed, commit hashes are the only safe way to handle this.

Neophytis:
> Never use versions tags on actions. Tags can be easily changed, commit hashes are the only safe way to handle this.

good catch, it's a good point it's a hardened way of working.

PR #442, opened 2026-06-22
Restore footer revision display in Docker production images

Neophytis:
## Summary
- Bake a `VERSION` file (short git SHA + commit timestamp) into production Docker images during CI build.
- Add `VersionInfo` utility to read `%kernel.project_dir%/VERSION` reliably, replacing the broken `../VERSION` path under FrankenPHP.
- Wire footer globals and feedback form version reporting through the shared utility.

## Context
Alpha stage runs `APP_ENV=prod` images that never created `VERSION` (excluded from build context; entrypoint only writes it in dev). The footer revision line went blank after the Docker deploy pipeline landed.

## Test plan
- [ ] Merge to `develop` and wait for CI + image build
- [ ] After stage auto-deploy, verify footer shows e.g. `rev. 2c1891a (2026-06-22 …)` on https://alpha.stage.bewelcome.org
- [ ] On the stage host: `docker compose … exec app cat /srv/bewelcome/VERSION` returns the 7-char short SHA
- [ ] Submit feedback and confirm the email includes the version string

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -71,6 +71,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # tag=v4.2.2
 
+      - name: Commit metadata for VERSION file
+        id: git
+        run: echo "commit_date=$(git log -1 --format=%aI)" >> "$GITHUB_OUTPUT"
+
       - name: Log in to GitHub Container Registry
         uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772  # tag=v3.4.0
         with:
@@ -97,6 +101,9 @@ jobs:
           target: bewelcome_php
           platforms: ${{ matrix.platform }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            GIT_SHA=${{ github.sha }}
+            GIT_COMMIT_DATE=${{ steps.git.outputs.commit_date }}
           cache-from: type=gha,scope=${{ matrix.arch }}
           cache-to: type=gha,mode=max,scope=${{ matrix.arch }}
           outputs: type=image,name=${{ env.IMAGE }},push-by-digest=true,name-canonical=true,push=true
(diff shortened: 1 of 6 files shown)


thisismeonmounteverest (changes requested):
Why not simply execute the following commands:

	git rev-parse --short HEAD > VERSION
	touch -d $(TIME_STAMP) VERSION

Neophytis:
## Lion compatibility — this does not break production

**Short answer:** Lion keeps working. Deployer already runs `make version` on every deploy (`deploy.php` → `make build version`), which executes exactly:

```bash
git rev-parse --short HEAD > VERSION
touch -d $(TIME_STAMP) VERSION
```

That is unchanged by this PR.

### What actually changed

| | **Lion (Deployer)** | **Alpha stage/prod (Docker)** |
|---|---|---|
| How `VERSION` is created | `make version` at deploy time (git checkout present) | Baked into the image at CI build time from `github.sha` |
| Where it lives | Project root (`{{release_path}}/VERSION`) | `/srv/bewelcome/VERSION` |
| How the app reads it | `%kernel.project_dir%/VERSION` via `VersionInfo` | Same |

The old code read `../VERSION`, which only worked when PHP's working directory happened to be `public/`. The new `VersionInfo` class reads the project-root file directly — the same path `make version` writes to on Lion. So Lion should behave the same or more reliably, not worse.

### Why not simply run `git rev-parse` in Docker?

Because production Docker images deliberately contain **no `.git` directory** (`.dockerignore` excludes it; the image is self-contained). There is nothing to `rev-parse` at build or runtime on stage.

The Dockerfile therefore reproduces the same result CI-side:

```dockerfile
printf '%s\n' "${GIT_SHA}" | cut -c1-7 > VERSION
touch -d "${GIT_COMMIT_DATE}" VERSION   # when provided
```

That is the Docker equivalent of `make version`, not a different mechanism.

Local dev (`APP_ENV != prod`) still runs `git rev-parse --short HEAD > VERSION` in the entrypoint when the file is missing — also unchanged.

Neophytis:
> Why not simply execute the following commands:
> 
> ```
> git rev-parse --short HEAD > VERSION
> touch -d $(TIME_STAMP) VERSION
> ```

Production version has no git checkout. the .git folde is removed in docker

Neophytis:
Stepping back on this one — shevek will handle the VERSION/footer approach. The Docker/CI direction (create VERSION at build time, copy into the image) can be picked up separately if needed.

Issue #381, opened 2025-10-31
Add templates for conversations


thisismeonmounteverest:
We'd like to add simple templates to the conversations systems that can be used for replies (not new messages as we do not want copy/paste requests). 

Templates should allow to use placeholder like username or telephone numbers but that's not necessary for the first version.

buyolitsez:
@thisismeonmounteverest, hi! what future directions do you see for this issue? adding templates directly to CKEditorToolbar?

thisismeonmounteverest:
Being able to address the member properly by name (short name or username) would be good. Allowing to select which templates to use (from a pre-defined list) might also be interesting.

And we need to see if the users disagree with the only for requests and invitations decision taken in the pull request #443.

Regarding positioning of the template dropdown: Not sure, I would wait for feedback. It would probably a good idea to collect telemetry data how often the templates are used and do an A/B test with different positions and compare the results.

PR #443, opened 2026-06-23
Add reply templates to conversation replies

buyolitsez:
This PR is an early draft of #381 

It adds simple reply templates to the existing conversation reply forms, with the template text inserted into CKEditor from a small dropdown.

I would like to talk about the vision behind those templates. I have added only three very basic templates. @thisismeonmounteverest, what do you think is the best way to extend them?

telegram-cloud-photo-size-2-5318922821637445585-w

telegram-cloud-photo-size-2-5318922821637445586-x

### DIFF
--- a/assets/js/roxeditor.js
+++ b/assets/js/roxeditor.js
@@ -46,7 +46,7 @@ function SpecialCharactersTextExtended( editor ) {
     ] );
 }
 
-function insertConversationTemplate( editor, template ) {
+function insertReplyTemplate( editor, template ) {
     if (!template) {
         return;
     }
@@ -57,7 +57,7 @@ function insertConversationTemplate( editor, template ) {
     editor.model.insertContent(modelFragment);
 }
 
-function registerConversationTemplateHandlers( editor ) {
+function registerReplyTemplateHandler( editor ) {
     const editorId = editor.sourceElement.id;
     const templateButtons = document.querySelectorAll('.js-conversation-template');
 
@@ -72,7 +72,7 @@ function registerConversationTemplateHandlers( editor ) {
 
         button.dataset.conversationTemplateRegistered = 'yes';
         button.addEventListener('click', () => {
-            insertConversationTemplate(editor, button.dataset.conversationTemplate);
+            insertReplyTemplate(editor, button.dataset.conversationTemplate);
         });
     });
 }
@@ -219,7 +219,9 @@ sourceElements.forEach( (element) => {
             const form = editor.sourceElement.closest('form')
 
             registerSubmitHandler(form);
-            registerConversationTemplateHandlers(editor);
+            if (editor.sourceElement.dataset.includeReplyTemplates === 'yes') {
+                registerReplyTemplateHandler(editor);
+            }
 
             const storedData = JSON.parse(window.localStorage.getItem(editor.sourceElement.id));
             if (storedData !== null) {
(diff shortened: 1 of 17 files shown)


assets/js/roxeditor.js
```diff
@@ -212,9 +244,13 @@ sourceElements.forEach( (element) => {
...
 function registerSubmitHandler( form ) {
-    form.addEventListener('submit', function( form ) {
+    if (!form) {
+        return;
+    }
+
+    form.addEventListener('submit', function() {
         // Remove data from localeStorage.
-        for (let [editor] of editors.entries()) {
+        for (let [, editor] of editors.entries()) {
```
thisismeonmounteverest: What is the comma doing here?

assets/js/roxeditor.js
```diff
@@ -212,9 +244,13 @@ sourceElements.forEach( (element) => {
 
 
 function registerSubmitHandler( form ) {
-    form.addEventListener('submit', function( form ) {
+    if (!form) {
```
thisismeonmounteverest: Did you run into a situation where this was the case or is this just defensive programming?

assets/js/roxeditor.js
```diff
@@ -46,6 +46,37 @@ function SpecialCharactersTextExtended( editor ) {
     ] );
 }
 
+function insertConversationTemplate( editor, template ) {
```
thisismeonmounteverest: Probably a better idea to add a include_templates option to the CKEditorType and add the reply templates dropdown in the CKeditor.twig. That option could then be check here before calling registerReplyTemplates.

assets/js/roxeditor.js
```diff
@@ -46,6 +46,37 @@ function SpecialCharactersTextExtended( editor ) {
...
+        return;
+    }
+
+    editor.editing.view.focus();
+    const viewFragment = editor.data.processor.toView(template);
+    const modelFragment = editor.data.toModel(viewFragment);
+    editor.model.insertContent(modelFragment);
+}
+
+function registerConversationTemplateHandlers( editor ) {
```
thisismeonmounteverest: This is only used for replies and I think we should keep it that way. Rename to registerReplyTemplateHandler.

thisismeonmounteverest (changes requested):
There's probably a way to add the templates dropdown to the CKEditorToolbar but this a really nice start.

assets/js/roxeditor.js
buyolitsez: I wanted to make it explicit, I will change to something like 
```
for (const editor of editors.values()) {
```

assets/js/roxeditor.js
buyolitsez: defensive, I will remove it

src/Controller/MessageController.php
```diff
@@ -141,7 +141,9 @@ public function reply(Request $request, Message $message): Response
         $replyMessage = new Message();
         $replyMessage->setSubject($message->getSubject());
 
-        $replyForm = $this->createForm(MessageToMemberType::class, $replyMessage);
+        $replyForm = $this->createForm(MessageToMemberType::class, $replyMessage, [
```
thisismeonmounteverest: Adding reply templates to messages isn't that interesting I think. Why would I respond with 'I need more details' to a message?

thisismeonmounteverest (changes requested):

thisismeonmounteverest (approved):

Issue #316, opened 2023-01-28
Missing email address check


gabriel-BW:
On signup there is a check about the email address (checking if it's already used, and sending a confirmation link to check if it's yours). But when you change it on your profile, there is no check at all... It would be appropriate to add the same checks, even if it means being logged out until a new mail confirmation is performed. If necessary, changing the email address could be moved out of the profile (in the same way as changing the password).
I'm bringing this up now, because we had a complaint from someone who claimed an account had been created in her name with real data (photo, etc) about her and her email address, but she didn't create it herself. And that's actually possible, if someone created the account with their own email and then changed it to hers. So she was unexpectedly receiving hosting requests... This can be done as a bad joke, or as revenge from an ex-lover or something similar.

buyolitsez:
should be closed, see #444

PR #444, opened 2026-06-23
Fix email change confirmation. #316

buyolitsez:
Fix of #316 

- Keep the current email active when a member requests an email change
- Store requested addresses in `NewEmail` and only promote them after confirmation
- Send the email-change confirmation link to the pending new address
- Redirect old `/editmyprofile...` self-edit URLs to the modern profile editor and harden the legacy callback against direct email writes
- Preserve compatibility for already-sent old `confirm_email` links by redirecting pending email changes to `change_email`

@thisismeonmounteverest please review it

### DIFF
--- a/build/members/members.ctrl.php
+++ b/build/members/members.ctrl.php
@@ -536,6 +536,13 @@ class MembersController extends RoxControllerBase
         if (isset($args->post)) {
             $vars = $this->cleanVars($args->post);
             $request = $args->request;
+            $rights = new MOD_right;
+            if (!($rights->hasRight('Admin') || $rights->hasRight('SafetyTeam')))
+            {
+                $loggedInMember = $this->model->getLoggedInMember();
+                $vars['memberid'] = $loggedInMember->getPKValue();
+                $vars['Email'] = $loggedInMember->Email;
+            }
             $errors = $this->model->checkProfileForm($vars);
 
             $uploadFailed = false;
@@ -565,11 +572,6 @@ class MembersController extends RoxControllerBase
                 // show form again
                 $mem_redirect->post = $vars;
                 return false;
-            }
-            $rights = new MOD_right;
-            if (!($rights->hasRight('Admin') || $rights->hasRight('SafetyTeam')))
-            {
-                $vars['memberid'] = $this->model->getLoggedInMember()->getPKValue();
             }
             $vars['member'] = $this->getMember($vars['memberid']);
             $vars = $this->model->polishProfileFormValues($vars);
--- a/build/members/members.ctrl.php
+++ b/build/members/members.ctrl.php
@@ -536,13 +536,6 @@ class MembersController extends RoxControllerBase
         if (isset($args->post)) {
             $vars = $this->cleanVars($args->post);
             $request = $args->request;
-            $rights = new MOD_right;
-            if (!($rights->hasRight('Admin') || $rights->hasRight('SafetyTeam')))
-            {
-                $loggedInMember = $this->model->getLoggedInMember();
-                $vars['memberid'] = $loggedInMember->getPKValue();
-                $vars['Email'] = $loggedInMember->Email;
-            }
             $errors = $this->model->checkProfileForm($vars);
 
             $uploadFailed = false;
@@ -572,6 +565,11 @@ class MembersController extends RoxControllerBase
                 // show form again
                 $mem_redirect->post = $vars;
                 return false;
+            }
+            $rights = new MOD_right;
+            if (!($rights->hasRight('Admin') || $rights->hasRight('SafetyTeam')))
+            {
+                $vars['memberid'] = $this->model->getLoggedInMember()->getPKValue();
             }
             $vars['member'] = $this->getMember($vars['memberid']);
             $vars = $this->model->polishProfileFormValues($vars);
--- a/src/Controller/AccountController.php
+++ b/src/Controller/AccountController.php
@@ -139,6 +139,13 @@ class AccountController extends AbstractController
             return $this->redirectToRoute('members_profile', ['username' => $member->getUsername()]);
         }
 
+        $newEmail = $member->getNewEmail();
+        if (null === $newEmail) {
+            $this->addTranslatedFlash('error', 'flash.profile.mail.already.confirmed');
+
+            return $this->redirectToRoute('members_profile', ['username' => $member->getUsername()]);
+        }
+
         $changeEmailAddressForm = $this->createForm(ConfirmEmailAddressFormType::class, [
             'registration_key' => $registrationKey,
         ]);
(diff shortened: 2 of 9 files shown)


build/members/members.ctrl.php
```diff
@@ -536,6 +536,13 @@ public function editMyProfileCallback($args, $action, $mem_redirect, $mem_resend
         if (isset($args->post)) {
             $vars = $this->cleanVars($args->post);
             $request = $args->request;
+            $rights = new MOD_right;
```
thisismeonmounteverest: The controller will be removed as soon as the new profile pages are ready. No need to make these changes.

src/Controller/AccountController.php
```diff
@@ -139,14 +139,27 @@ public function changeEmailAddress(Request $request, Member $member, ?string $re
...
+
         $changeEmailAddressForm = $this->createForm(ConfirmEmailAddressFormType::class, [
             'registration_key' => $registrationKey,
         ]);
         $changeEmailAddressForm->handleRequest($request);
         if ($changeEmailAddressForm->isSubmitted() && $changeEmailAddressForm->isValid()) {
             $registrationKey = $changeEmailAddressForm['registration_key']->getData();
             if ($registrationKey === $member->getRegistrationKey()) {
-                $member->setEmail($member->getNewEmail());
+                if (!$this->signupModel->checkEmailAddress($newEmail, $member)) {
```
thisismeonmounteverest: This situation shouldn't arise: When the user applies for a new email any signup with the same email should/will be blocked. Why this extra precaution here?

src/Controller/ProfileController.php
```diff
@@ -149,6 +149,24 @@ public function redirectToSetLocation(): RedirectResponse
         return $this->redirectToRoute('profile_set_location', ['username' => $this->getUser()->getUsername()]);
     }
 
+    #[Route(path: '/editmyprofile', name: 'legacy_edit_profile_redirect', priority: 100)]
```
thisismeonmounteverest: This can be combined into a fetch all for brevity.

src/Controller/ProfileController.php
```diff
@@ -149,6 +149,24 @@ public function redirectToSetLocation(): RedirectResponse
...
+    #[Route(
+        path: '/editmyprofile/{legacyPath}',
+        name: 'legacy_edit_profile_path_redirect',
+        requirements: ['legacyPath' => '.+'],
+        priority: 100
+    )]
+    public function redirectLegacyEditProfile(): RedirectResponse
+    {
+        /** @var Member|null $member */
+        $member = $this->getUser();
```
thisismeonmounteverest, line 162: This can't happen as /editmyprofile is behind the firewall. There is always a user here.

src/Controller/SignupController.php
```diff
@@ -188,6 +188,13 @@ public function confirmEmailAddress(Request $request, Member $member, ?string $r
             return $this->redirectToRoute('members_profile', ['username' => $member->getUsername()]);
         }
 
+        if (null !== $member->getNewEmail()) {
```
thisismeonmounteverest: Would make more sense to me to block changing email if the first email address isn't confirmed yet.

src/Controller/SignupController.php
```diff
@@ -188,6 +188,13 @@ public function confirmEmailAddress(Request $request, Member $member, ?string $r
...
+            return $this->redirectToRoute('change_email', [
```
thisismeonmounteverest: The registration key should be either coming from a direct link or from user entry. This way the active confirmation wouldn't be enforced. But with the comment above this would be removed anyway.

src/Model/ProfileModel.php
```diff
@@ -224,7 +224,7 @@ public function sendEmailConfirmationEmail(Member $member, string $email): void
             'key' => $member->getRegistrationKey(),
         ];
 
-        $this->mailer->sendSignupEmail($member, 'newemail', $parameters);
+        $this->mailer->sendSignupEmailToAddress($member, $email, 'newemail', $parameters);
```
thisismeonmounteverest: Email address is already in the parameters. The new method wouldn't need $email and 'newmail' parameters in the signature.

src/Model/SignupModel.php
```diff
@@ -194,20 +194,22 @@ public function checkUsername(string $username): bool
     /**
      * Email is either taken directly or is currently changed and awaiting confirmation.
      */
-    public function checkEmailAddress(string $email): bool
+    public function checkEmailAddress(string $email, ?Member $ignoredMember = null): bool
```
thisismeonmounteverest: This looks weird. Does this function still has only one job to do?

src/Service/Mailer.php
```diff
@@ -108,6 +108,17 @@ public function sendSignupEmail(Member $receiver, string $template, $parameters)
         );
     }
 
+    public function sendSignupEmailToAddress(Member $receiver, string $emailAddress, string $template, $parameters): bool
+    {
+        return $this->sendTemplateEmail(
+            self::SIGNUP_EMAIL_ADDRESS,
+            new Address($emailAddress, $receiver->getUsername()),
```
thisismeonmounteverest: Use email address from parameters. set template name to 'newemail'.

src/Service/Mailer.php
```diff
@@ -277,21 +288,25 @@ private function getBeWelcomeAddress(Member $sender, $email): Address
      * @param Member|Address        $receiver
      * @param mixed                 $parameters
      */
-    private function sendTemplateEmail($sender, $receiver, string $template, array $parameters): bool
+    private function sendTemplateEmail($sender, $receiver, string $template, array $parameters, ?Member $receiverContext = null): bool
```
thisismeonmounteverest: Set receiver to Address in sendSignupEmailToAddress and the $receiverContext isn't necessary.

src/Service/Mailer.php
```diff
@@ -277,21 +288,25 @@ private function getBeWelcomeAddress(Member $sender, $email): Address
...
-            $locale = $receiver->getLocale();
-            $parameters['receiver'] = $receiver;
+            $receiverContext = $receiver;
             $receiver = new Address($receiver->getEmail(), $receiver->getUsername());
         } elseif (!$receiver instanceof Address) {
             $message = \sprintf('$receiver must be an instance of %s or %s.', Member::class, Address::class);
             throw new InvalidArgumentException($message);
         }
 
+        if ($receiverContext instanceof Member) {
```
thisismeonmounteverest: Might make more sense to set the locale from the outside instead of the receiverContext.

tests/Controller/AccountControllerTest.php
```diff
@@ -0,0 +1,254 @@
+<?php
```
thisismeonmounteverest, line 1: Nice an integration test along with the changes.

thisismeonmounteverest (changes requested):

buyolitsez:
@thisismeonmounteverest thanks for the review! I tried to fix all your comments

thisismeonmounteverest (approved):

PR #447, opened 2026-06-26
Embed build version metadata in production image

buyolitsez:
Fix production Docker image metadat to include a valid `VERSION` file in the deploy


buyolitsez:
@thisismeonmounteverest hi, could you please review it?

PR #448, opened 2026-06-26
Fix getactive inline contact markup

buyolitsez:
Fix of https://www.bewelcome.org/group/570/forum/s33699-html_or_not_in_translated_strings


buyolitsez:
@thisismeonmounteverest not sure whether I should tag you each time I create a PR, but please review it when you have time

thisismeonmounteverest (approved):
That's a straight forward solution for the GetActive page.

PR #449, opened 2026-06-27
Implement trip notifications and read state

### LINKED ISSUE #340: Add notifications for trips and allow to mark trips as 'read'
The trips feature at the moment isn't working as well as it could. One problem is that a lot of people do not use the landing page (as they seldom login) and another is that even if they do they always see the same trips as there is no way to remove trips of visitors.

Adding the following hopefully mitigates this:

* Send a notification when new trips are added in your area
    * Possible ways to do that: Immediately, once a day, once a week?
* Add way to hide a trip from the list of visitors, so that a new trip is shown instead (if any)

Notifications should only send as long as a member wants them, so an unsubscribe options is needed (can also be a preference). Unhiding a trip isn't needed but could be done on the general trips overview page.

buyolitsez:
To close #340 

- notify nearby hosts when a member creates a new public trip
- add a trip notification preference during signup/preferences
- allow hosts to hide trips from their visitors lists
- mark matching trip notifications as checked when a trip is hidden

telegram-cloud-photo-size-2-5330044562141026979-y

telegram-cloud-photo-size-2-5330044562141026982-y

telegram-cloud-photo-size-2-5330044562141027003-y


thisismeonmounteverest (changes requested):
@buyolitsez Looks good. There are a few problems, though.

The notification needs to be on the subtrip as the reason for 'reading' a trip might be either the person it self or the time of the visit. 

The notification tab is a left over from far, far away and will be removed in the near future so we do not need to worry about that anymore.

The original discussion was to allow for digests: Immediately, daily, weekly, monthly. Starting with immediately is fine for me, but with that in mind the code might need some tinkling.

To find the members to be informed we should make use of the new spatial indices for geonames, shouldn't we?

Do we want to allow members to 'read' a leg on the visitors page (/visitors) or only on the landing page?

Migrations/Version20260626100000.php
```diff
@@ -11,20 +11,22 @@ final class Version20260626100000 extends AbstractMigration
...
-        $this->addSql("INSERT INTO preferences (position, codeName, codeDescription, Description, created, DefaultValue, PossibleValues, Status) SELECT 65, 'TripsNotifications', 'trips.notifications', 'How often the member wants notifications for trips in their area', NOW(), 'No', 'No;Yes', 'Normal' WHERE NOT EXISTS (SELECT 1 FROM preferences WHERE codeName = 'TripsNotifications')");
+        $this->addSql('CREATE TABLE IF NOT EXISTS member_subtrips_read (id INT AUTO_INCREMENT NOT NULL, member_id INT NOT NULL, subtrip_id INT NOT NULL, created DATETIME NOT NULL, INDEX IDX_B1EC0277597D3FE (member_id), INDEX IDX_B1EC027F2BD7DD7 (subtrip_id), UNIQUE INDEX member_subtrip_read_unique (member_id, subtrip_id), CONSTRAINT FK_B1EC0277597D3FE FOREIGN KEY (member_id) REFERENCES member (id) ON DELETE CASCADE, CONSTRAINT FK_B1EC027F2BD7DD7 FOREIGN KEY (subtrip_id) REFERENCES sub_trips (id) ON DELETE CASCADE, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB');
```
thisismeonmounteverest: Doctrine prefers singluar table names. Use member_subtrip_read accordingly.

Migrations/Version20260626100000.php
```diff
@@ -11,20 +11,22 @@ final class Version20260626100000 extends AbstractMigration
...
+        $this->addSql("UPDATE memberspreferences mp INNER JOIN preferences p ON p.id = mp.IdPreference SET mp.Value = 'Immediately' WHERE p.codeName = 'TripsNotifications' AND mp.Value = 'Yes'");
     }
 
     #[Override]
     public function down(Schema $schema): void
     {
         $this->addSql("DELETE mp FROM memberspreferences mp INNER JOIN preferences p ON p.id = mp.IdPreference WHERE p.codeName = 'TripsNotifications'");
         $this->addSql("DELETE FROM preferences WHERE codeName = 'TripsNotifications'");
-        $this->addSql('DROP TABLE IF EXISTS member_trips_read');
+        $this->addSql('DROP TABLE IF EXISTS member_subtrips_read');
```
thisismeonmounteverest: member_subtrip_read

src/Controller/TripController.php
```diff
@@ -180,16 +180,16 @@ public function copy(Trip $trip): Response
         return $this->redirectToRoute('trip_edit', ['id' => $newTrip->getId()]);
     }
 
-    #[Route(path: '/trip/{id}/read', name: 'trip_mark_read', requirements: ['id' => '\d+'], methods: ['POST'])]
-    public function markRead(Request $request, Trip $trip): RedirectResponse
+    #[Route(path: '/trip/leg/{id}/read', name: 'trip_mark_read', requirements: ['id' => '\d+'], methods: ['POST'])]
+    public function markRead(Request $request, Subtrip $subtrip): RedirectResponse
     {
-        if (!$this->isCsrfTokenValid('trip_read' . $trip->getId(), $request->request->get('_token'))) {
+        if (!$this->isCsrfTokenValid('subtrip_read' . $subtrip->getId(), $request->request->get('_token'))) {
```
thisismeonmounteverest: Isn't this normally done within symfony/forms? Why calling it explicitily?

src/Entity/MemberSubtripRead.php
```diff
@@ -5,19 +5,19 @@
 use DateTime;
 use Doctrine\ORM\Mapping as ORM;
 
-#[ORM\Table(name: 'member_trips_read')]
-#[ORM\UniqueConstraint(name: 'member_trip_read_unique', columns: ['member_id', 'trip_id'])]
+#[ORM\Table(name: 'member_subtrips_read')]
```
thisismeonmounteverest: The better wording would probably be 'hide' instead of 'read'. That would push the table name to be member_subtrip_hidden..

src/Entity/Preference.php
```diff
@@ -36,7 +36,11 @@ class Preference
     public const TRIPS_VICINITY_RADIUS = 'TripLegsVicinityRadius';
     public const TRIP_NOTIFICATIONS = 'TripsNotifications';
     public const TRIP_NOTIFICATIONS_NEVER = 'No';
-    public const TRIP_NOTIFICATIONS_IMMEDIATELY = 'Yes';
+    public const TRIP_NOTIFICATIONS_IMMEDIATELY = 'Immediately';
+    public const TRIP_NOTIFICATIONS_LEGACY_IMMEDIATELY = 'Yes';
```
thisismeonmounteverest: Unnecessary.

src/Repository/SubtripRepository.php
```diff
@@ -79,72 +79,117 @@ public function getMembersToNotifyAboutTrip(Trip $trip, int $duration = 3): arra
...
+                ->setParameter('creator', $trip->getCreator())
+                ->setParameter('travellers', $trip->getCountOfTravellers())
+                ->setParameter('tripNotificationPreference', $tripNotificationPreference)
+                ->setParameter('radiusPreference', $radiusPreference)
+                ->setParameter('defaultNotification', Preference::TRIP_NOTIFICATIONS_NEVER)
+                ->setParameter(
+                    'immediateNotifications',
+                    [
+                        Preference::TRIP_NOTIFICATIONS_IMMEDIATELY,
+                        Preference::TRIP_NOTIFICATIONS_LEGACY_IMMEDIATELY,
```
thisismeonmounteverest: Remove here as well.

translations/missing/general.yaml
```diff
@@ -74,7 +74,7 @@ trips.never:
     - One of the options for trips notifications (legacy value).
 trips.immediately:
```
thisismeonmounteverest, line 69: Remove.

translations/missing/preference.yaml
```diff
@@ -69,5 +69,14 @@ tripsnotificationsnever:
     - '@no'
     - Option on the preferences page for trip notifications (legacy value).
```
thisismeonmounteverest: That's not a legacy value right? You can still say you do not want any notifications.

translations/missing/trips.yaml
```diff
@@ -128,9 +128,6 @@ trip.show:
 trip.mark.read:
```
thisismeonmounteverest: Strong argument to rename that functionality...

thisismeonmounteverest (changes requested):

Migrations/Version20260626100000.php
```diff
@@ -16,17 +16,18 @@ public function getDescription(): string
...
+        $this->addSql('CREATE TABLE IF NOT EXISTS member_trip_notification_sent (id INT AUTO_INCREMENT NOT NULL, member_id INT NOT NULL, trip_id INT NOT NULL, created DATETIME NOT NULL, INDEX IDX_9D9311917597D3FE (member_id), INDEX IDX_9D931191A5BC2E0E (trip_id), UNIQUE INDEX member_trip_notification_sent_unique (member_id, trip_id), CONSTRAINT FK_9D9311917597D3FE FOREIGN KEY (member_id) REFERENCES member (id) ON DELETE CASCADE, CONSTRAINT FK_9D931191A5BC2E0E FOREIGN KEY (trip_id) REFERENCES trips (id) ON DELETE CASCADE, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB');
         $this->addSql("INSERT INTO preferences (position, codeName, codeDescription, Description, created, DefaultValue, PossibleValues, Status) SELECT 65, 'TripsNotifications', 'trips.notifications', 'How often the member wants notifications for trips in their area', NOW(), 'No', 'No;Immediately;Daily;Weekly;Monthly', 'Normal' WHERE NOT EXISTS (SELECT 1 FROM preferences WHERE codeName = 'TripsNotifications')");
         $this->addSql("UPDATE preferences SET PossibleValues = 'No;Immediately;Daily;Weekly;Monthly' WHERE codeName = 'TripsNotifications'");
```
thisismeonmounteverest: This would only change the database if the previous statement didn't. For new setups that will always be the case (the situation that needs cleanup can only arise during development).

src/Command/SendTripNotificationsCommand.php
```diff
@@ -0,0 +1,59 @@
...
+        $frequency = strtolower((string) $input->getArgument('frequency'));
+
+        if (!isset(self::FREQUENCIES[$frequency])) {
+            $io->error('Frequency must be daily, weekly or monthly.');
+
+            return Command::INVALID;
+        }
+
+        [$preferenceValue, $period] = self::FREQUENCIES[$frequency];
+        $until = new DateTimeImmutable();
```
thisismeonmounteverest: I always thought about trip notifications as information about upcoming trips. This seems to target trips in the past.

src/Entity/MemberTripNotificationSent.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace App\Entity;
+
+use DateTime;
+use Doctrine\ORM\Mapping as ORM;
+
+#[ORM\Table(name: 'member_trip_notification_sent')]
```
thisismeonmounteverest: As you hide legs of a trip notifications should be sent for legs as well, shouldn't they?

src/Model/TripModel.php
```diff
@@ -151,25 +159,87 @@ public function markSubtripAsRead(Member $member, Subtrip $subtrip): void
     }
 
     public function notifyHostsAboutTrip(Trip $trip): void
```
thisismeonmounteverest, line 161: The important information is that there is a traveller visiting your area. Which means we probably want to notifyHostsAboutVisitors(SubTrip $leg).

src/Model/TripModel.php
```diff
@@ -151,25 +159,87 @@ public function markSubtripAsRead(Member $member, Subtrip $subtrip): void
...
+                ++$sentCount;
+            } finally {
+                $this->releaseTripNotificationLock($lockName);
+            }
         }
+
+        return $sentCount;
+    }
+
+    private function acquireTripNotificationLock(Member $member, Trip $trip): ?string
```
thisismeonmounteverest: Which part of the database is locked here? If more than the member_leg_notification_sent table this might impact the whole site.

src/Entity/Preference.php
```diff
@@ -37,7 +37,6 @@ class Preference
     public const TRIP_NOTIFICATIONS = 'TripsNotifications';
     public const TRIP_NOTIFICATIONS_NEVER = 'No';
```
thisismeonmounteverest: Should be Never.

tests/Model/CreatedTripsRepository.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
thisismeonmounteverest: I'd expect all things trips to be in the same repository.

tests/Model/SentTripNotificationsRepository.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
thisismeonmounteverest: This doesn't even use the database, right? So why is it a repository?

tests/Model/TripModelNotificationTest.php
```diff
@@ -88,16 +92,26 @@ public function testNotifyHostsAboutTripSendsEmailsToMatchingHostsOnce(): void
         $this->setEntityId($trip, 42);
 
         $subtripRepository = $this->createSubtripRepository([$host, $sameHost]);
+        $sentRepository = new SentTripNotificationsRepository();
 
         $entityManager = $this->createMock(EntityManagerInterface::class);
         $entityManager
-            ->expects($this->once())
+            ->expects($this->exactly(2))
```
thisismeonmounteverest: This binds the test against implementation which isn't necessary.

tests/Model/TripModelNotificationTest.php
```diff
@@ -155,8 +331,22 @@ private function setEntityId(object $entity, int $id): void
         $property->setValue($entity, $id);
     }
 
-    private function createSubtripRepository(array $hosts): EntityRepository
+    private function createSubtripRepository(array $hosts): MatchingHostsRepository
```
thisismeonmounteverest: A MatchingHostsRepository is not a SubtripRepository.

thisismeonmounteverest (changes requested):

src/Model/TripModel.php
buyolitsez: I changed it to be scoped to key member-subtrip-notification-sent:{memberId}:{subtripId}. So it only prevents two workers from sending the same notification to the same member for the same visitor leg at the same time.

thisismeonmounteverest (approved):

PR #450, opened 2026-07-11
Feature/168 browser web push

### LINKED ISSUE #168: Add browser notificcation via webpush
There is no plan for now to create an app. But, phone web browser can send notification to the phone or desktop via [webpush](https://support.mozilla.org/en-US/kb/push-notifications-firefox)
It would be great to send some notification to user about host request / private messages or thread subscribed.

I talked about it [here](https://www.bewelcome.org/forums/s21005/#post105810)

buyolitsez:
To close #168 

- Add browser notifications for private messages, hosting requests, invitations, and subscribed forum threads.
- Support three preference modes:
  - `No`: notifications disabled.
  - `OpenOnly`: notifications displayed while BeWelcome is open.
  - `Always`: Web Push delivery with BeWelcome tabs closed.

Deployment notes:

- Configure valid `WEB_PUSH_VAPID_SUBJECT`, `WEB_PUSH_VAPID_PUBLIC_KEY`, and `WEB_PUSH_VAPID_PRIVATE_KEY` values.
- Schedule `bin/console send:notifications` frequently.
- Schedule `bin/console data:retention` daily.
- Import the new entries from `translations/missing`.

image


src/Command/DataRetentionCommand.php
thisismeonmounteverest, line 1: One responsibility principle says not to add the cleanup of brower psuh data into the member cleanup command.

src/Command/SendAndPushNotificationsCommand.php
thisismeonmounteverest, line 1: Rename to SendAndPushNotificationsCommand?

src/Command/SendAndPushNotificationsCommand.php
```diff
@@ -38,6 +43,9 @@ public function __construct(
         private EntityManagerInterface $entityManager,
         private LoggerInterface $logger,
         private Mailer $mailer,
+        private BrowserNotificationService $browserNotificationService,
```
thisismeonmounteverest, line 46: Shouldn't these be the Interfaces instead of the real thing? Otherwise I do not see why the interfaces were introduced in the first place.

src/Controller/BrowserPushSubscriptionController.php
```diff
@@ -0,0 +1,259 @@
...
+    {
+        $subscriptions = $this->entityManager->getRepository(BrowserPushSubscription::class)->findBy(
+            ['member' => $member],
+            ['lastSeen' => 'DESC', 'id' => 'DESC']
+        );
+        if (\count($subscriptions) <= self::MAX_SUBSCRIPTIONS_PER_MEMBER) {
+            return;
+        }
+
+        foreach (\array_slice($subscriptions, self::MAX_SUBSCRIPTIONS_PER_MEMBER) as $subscription) {
```
thisismeonmounteverest: Doesn't this just remove MAX_SUBSCRIPTIONS_PER_MEMBER from the subscriptions instead of keeping that number?

src/Controller/PasswordController.php
thisismeonmounteverest, line 1: What#s this change about?

src/Model/MemberDataExtractor/BrowserPushNotificationDeliveriesExtractor.php
thisismeonmounteverest, line 1: Aren't we cleaning up on the browser pushes? Why extract them?

thisismeonmounteverest (changes requested):

src/Command/SendAndPushNotificationsCommand.php
buyolitsez: There were no interfaces for those services, and the existing interfaces do not expose their `queue()` or `process()` operations. So, since each service currently has only one implementation, I think that introducing a new interface would be strange
Do you think we should add a new interface?

src/Model/MemberDataExtractor/BrowserPushNotificationDeliveriesExtractor.php
buyolitsez: I made the cleanup time-based - when `browser-push:retention` runs, notifications older than 7 days are deleted together with their delivery records, and inactive subscriptions older than one year are also removed

Until then, everything remains stored and linked to the member (recent or pending notifications, their delivery attempts/status/errors, and active subscriptions). Therefore, they should be included in the data export

src/Command/SendAndPushNotificationsCommand.php
thisismeonmounteverest: No, I do not think we need an interface here. There isn't the need for further abstraction. The reason for my comment was that you introduced the interface but didn't use it afterwards.

src/Model/MemberDataExtractor/BrowserPushNotificationDeliveriesExtractor.php
thisismeonmounteverest: Okay. Why do we keep the data for 7 days then? Isn't 1 enough and therefore doesn't need to be extracted?

thisismeonmounteverest (approved):

PR #452, opened 2026-07-13
Fix forum notification email threading

### LINKED ISSUE #288: Notification emails are not threaded
The notifications BW sends from a group/forum/etc. discussion are not threaded, they probably don't use a standard email thread ID... so email clients can't properly group/thread the notification messages of the same conversation.

Please add the threading metadata to the headers of the notification emails being sent out by BW.

buyolitsez:
Fixes #288

New-thread notifications now use a specific Message-ID

### DIFF
--- a/src/Service/Mailer.php
+++ b/src/Service/Mailer.php
@@ -144,24 +144,14 @@ class Mailer
 
     public function sendNotificationEmail(Address $sender, Member $receiver, $parameters): bool
     {
-        $notification = $parameters['notification'];
-        $threadMessageId = \sprintf(
-            'forum-thread-%d-member-%d%s',
-            $notification->getPost()->getThread()->getId(),
-            $receiver->getId(),
-            strrchr($sender->getAddress(), '@')
-        );
-        $identificationHeaders = 'newthread' === $notification->getType()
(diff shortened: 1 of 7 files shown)


src/Service/Mailer.php
```diff
@@ -144,11 +144,31 @@ public function sendNewsletterEmail(Newsletter $newsletter, Member $receiver, ar
 
     public function sendNotificationEmail(Address $sender, Member $receiver, $parameters): bool
     {
+        $notification = $parameters['notification'];
```
thisismeonmounteverest: Unfortunately, this is wrong according to RFC 2822: Each notification would need to list all parents which will get really cumbersome with the way forum threads work. 

   The "In-Reply-To:" field will contain the contents of the "Message-
   ID:" field of the message to which this one is a reply (the "parent
   message").  If there is more than one parent message, then the "In-
   Reply-To:" field will contain the contents of all of the parents'
   "Message-ID:" fields.  If there is no "Message-ID:" field in any of
   the parent messages, then the new message will have no "In-Reply-To:"
   field.

thisismeonmounteverest (changes requested):

thisismeonmounteverest (approved):

PR #453, opened 2026-07-13
Delete groups with all memberships

### LINKED ISSUE #80: Groups are deleted but membership isn't
When a group gets deleted the code should remove the membership entries in the database as well but obviously fails to do so. This wasn't a problem with the old code but the new code uses the memberships as the entry point and not the groups so this leads to problems.

A database cleanup is needed and the group deletion process needs to be changed so that membership is really removed. (Use cascade in Doctrine entity class?)

buyolitsez:
Clean up orphaned memberships and report deletion failures.

Fixes #80


thisismeonmounteverest:
Instead of fixing this in the legacy code it would be better to add that to the 'new' group controller and model in src/Controller and src/Model.

tests/Legacy/GroupDeleteDouble.php
thisismeonmounteverest, line 1: Should be GroupsDeleteDoubleTest.php

src/Controller/GroupController.php
```diff
@@ -315,6 +316,38 @@ public function withdrawInviteMemberGroup(Request $request, Group $group, Member
...
+    #[Route(
+        path: '/group/{id}/delete/true',
+        name: 'group_delete_confirmed',
+        requirements: ['id' => '\\d+'],
+        methods: ['GET'],
+        priority: 10,
+    )]
+    public function deleteGroup(Group $group, Logger $logger): RedirectResponse
+    {
+        /** @var Member|null $member */
```
thisismeonmounteverest: Member will never be null here as the route is behind the firewall.

src/Controller/GroupController.php
```diff
@@ -315,6 +316,38 @@ public function withdrawInviteMemberGroup(Request $request, Group $group, Member
...
+        $member = $this->getUser();
+        $isGroupsAdmin = $member instanceof Member
```
thisismeonmounteverest: First part of conditional is always true.

src/Controller/GroupController.php
```diff
@@ -315,6 +316,38 @@ public function withdrawInviteMemberGroup(Request $request, Group $group, Member
...
+            && $this->isGranted(Member::ROLE_ADMIN_GROUP)
+            && 10 === $member->getLevelForRight(Member::ROLE_ADMIN_GROUP);
+        if (!$member instanceof Member || (!$group->isAdmin($member) && !$isGroupsAdmin)) {
+            throw $this->createAccessDeniedException('No group admin');
```
thisismeonmounteverest: Not sure if this is shown to the user but if it is it should be translated.

tests/Controller/GroupControllerTest.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $client = static::createClient();
+        $entityManager = static::getContainer()->get(EntityManagerInterface::class);
+        $connection = $entityManager->getConnection();
+        $owner = $this->getMember($entityManager, 'member-2');
+        $applicant = $this->getMember($entityManager, 'member-1');
+        $inactiveMember = $this->getMember($entityManager, 'member-banned');
+        $groupId = random_int(1_500_000_000, 2_000_000_000);
+
+        try {
```
thisismeonmounteverest: While I understand this defensive programming style in a test I would not expect that any exception would happen if I execute the method under test.

What's the verdict of the test if the createGroupFixture call fails?

The asserts should always be executed. So if the try block is necessary move the asserts into the finally section.

tests/Controller/GroupControllerTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+    private function getMember(EntityManagerInterface $entityManager, string $username): Member
+    {
+        $member = $entityManager->getRepository(Member::class)->findOneBy(['username' => $username]);
+        self::assertInstanceOf(Member::class, $member);
+
+        return $member;
+    }
+
+    private function createGroupFixture(
```
thisismeonmounteverest, line 66: Aren't we still using dama/doctrine-test-bundle? So adding and removing would be part of the test setup and reverted without the need of a finally block.

tests/Controller/GroupControllerTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+    private function countRows(Connection $connection, string $table, string $column, int $groupId): int
+    {
+        return (int) $connection->fetchOne(
+            "SELECT COUNT(*) FROM `{$table}` WHERE `{$column}` = ?",
+            [$groupId],
+        );
+    }
+
+    private function removeGroupFixture(Connection $connection, int $groupId): void
```
thisismeonmounteverest: Let dama/doctrine-test-bundle handle this.

thisismeonmounteverest (changes requested):

PR #456, opened 2026-07-15
Feature/168 browser web push

buyolitsez:


config/services.yaml
thisismeonmounteverest, line 88: These interfaces aren't used anywhere, are they?

src/Controller/BrowserPushSubscriptionController.php
thisismeonmounteverest, line 1: Where is this controller used?

config/services.yaml
buyolitsez: They are used as autowiring contracts: `PushGatewayInterface` is injected into `BrowserNotificationService`, `BrowserPushEndpointResolverInterface` into `BrowserPushEndpointValidator`, and `BrowserPushWebPushFactoryInterface` into `PushGateway`

src/Controller/BrowserPushSubscriptionController.php
buyolitsez: in post/delete endpoints used by `browserPushPreference.js`. The preference page passes these route urls to the script, which calls the controller to manage the server-side subscription

config/services.yaml
thisismeonmounteverest: If the interface is just the class why not use the class instead?

thisismeonmounteverest: But I merge it anyway would like to see it in action :)

PR #457, opened 2026-07-18
Fix admin rights member lists

### LINKED ISSUE #154: Admin righs list - choose a user ends up in error page and unable to assign a right
https://www.bewelcome.org/admin/rights/list/members

Choosing my username an click on "refresh" button I get an error page. Also to initial display of the list is quite heavy and takes awhile.

buyolitsez:
To close #154 

Maybe some more optimizations should be done to fix "Also to initial display of the list is quite heavy and takes awhile."


thisismeonmounteverest:
Moving this (and flags) to new code would make more sense.

Removing the history isn't an option, though, as it is important to know if someone had the rights assigned at some point in time. Or to re-assign it.

That placename and country were never used is an oversight that needs to be mitigated not enforced :)

thisismeonmounteverest (approved):

PR #468, opened 2026-08-03
fix: use noreply@ as From address for forum and group notifications

Neophytis:
## Summary

`forum@bewelcome.org` and `group@bewelcome.org` were hardcoded as the `From` address in `SendNotificationsCommand::determineSender()`. Member replies to these notifications were landing in the OTRS raw queue.

Changed both to `noreply@bewelcome.org`, which has a Mailcow Sieve auto-reply configured that responds: *"This mailbox is not monitored."*

## Changes

1 file, 2 lines changed in `src/Command/SendNotificationsCommand.php`.

## Test plan

- [ ] Trigger a forum post notification and confirm `From` header shows `noreply@bewelcome.org`
- [ ] Reply to a notification and confirm auto-reply is received

### DIFF
--- a/src/Command/SendNotificationsCommand.php
+++ b/src/Command/SendNotificationsCommand.php
@@ -123,9 +123,9 @@ class SendNotificationsCommand extends Command
     {
         $thread = $post->getThread();
         if ($thread->getGroup()) {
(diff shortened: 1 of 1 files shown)


Neophytis:
Closing to investigate master properly before opening a PR.

PR #470, opened 2026-08-04
fix(docker): fix 5 bugs in Shevek's PHP 8.2 Docker setup

Neophytis:
## Summary

Fixes 5 bugs introduced in commit 901b23a9c ("Make master work on php8.2 and update docker setup accordingly"):

- **Security**: Remove DB credential echo from entrypoint — the debug `echo "new PDO(...)"` line printed the full database password to `docker logs` on every container start
- **Production DOA**: Add env-var fallbacks after `.env` grep — `composer dump-env prod` deletes `.env` in the production image, causing all five credential variables to be empty strings; the PDO wait loop then fails for 300s and exits 1, making production containers unbootable
- **Migrations silently skipped**: Fix path `src/Migrations/` → `Migrations/` in entrypoint — `src/Migrations/` does not exist; the condition always evaluated false
- **Migrations absent from image**: Add `COPY Migrations Migrations/` to Dockerfile — migration files were not copied into the production image at all
- **CI testing wrong PHP version**: Update all 11 jobs from `php-version: 7.4` to `8.2`, and the Behat MariaDB service from `10.1.41` to `12.0.2` to match the Dockerfile

## Test plan

- [ ] Build the production image: `docker build --target bewelcome_php -t bewelcome/app .`
- [ ] Verify no credentials appear in container logs: `docker logs <container>` should not contain any PDO connection strings
- [ ] Verify migrations run on a fresh container start with `APP_ENV=prod` and DB credentials passed as env vars (not via `.env`)
- [ ] Verify `Migrations/` directory is present in the built image: `docker run --rm --entrypoint sh bewelcome/app -c 'ls Migrations/'`
- [ ] CI passes on PHP 8.2

### DIFF
--- a/docker/php/docker-entrypoint.sh
+++ b/docker/php/docker-entrypoint.sh
@@ -53,11 +53,15 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 	database_name=$(grep '^DB_NAME=' .env | cut -f 2 -d '=')
 	database_user=$(grep '^DB_USER=' .env | cut -f 2 -d '=')
 	database_password=$(grep '^DB_PASS=' .env | cut -f 2 -d '=')
+	database_host="${database_host:-${DB_HOST:-db}}"
+	database_port="${database_port:-${DB_PORT:-3306}}"
+	database_name="${database_name:-${DB_NAME:-bewelcome}}"
+	database_user="${database_user:-${DB_USER:-bewelcome}}"
+	database_password="${database_password:-${DB_PASSWORD:-bewelcome}}"
 
 	echo "Waiting for db to be ready..."
 	db_wait_seconds=0
 	db_wait_max="${BEWELCOME_DB_READY_MAX_WAIT_SECONDS:-300}"
-	echo "new PDO('mysql:host=${database_host};port=${database_port};dbname=${database_name}', '${database_user}', '${database_password}');"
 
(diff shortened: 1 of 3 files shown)


docker/php/docker-entrypoint.sh
```diff
@@ -53,11 +53,15 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 	database_name=$(grep '^DB_NAME=' .env | cut -f 2 -d '=')
 	database_user=$(grep '^DB_USER=' .env | cut -f 2 -d '=')
 	database_password=$(grep '^DB_PASS=' .env | cut -f 2 -d '=')
+	database_host="${database_host:-${DB_HOST:-db}}"
```
thisismeonmounteverest, line 56: This overrides the information grepped from the .env file just above, doesn't it?

docker/php/docker-entrypoint.sh
```diff
@@ -81,7 +85,7 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		if [ -f docker/db/geonamesadminunits.sql ]; then
 			mysql $database_name -u $database_user -p$database_password -h $database_host < docker/db/geonamesadminunits.sql
 		fi
-	elif ls -A src/Migrations/*.php > /dev/null 2>&1; then
+	elif ls -A Migrations/*.php > /dev/null 2>&1; then
```
thisismeonmounteverest, line 88: Master will use the production database on deployments no migrations needed at all.

That also means for master we might want to have a docker setup that stops when APP_ENV isn't set to prod.

thisismeonmounteverest (approved):

PR #472, opened 2026-08-05
fix(mailer): remove freemail Reply-To to fix rspamd SPOOF_REPLYTO junking

Neophytis:
## Problem

PR #471 fixed DMARC by switching to `From: noreply@bewelcome.org`, but kept `Reply-To: <reporter@freemail>` so staff could reply directly. This still causes feedback emails to land in Junk — rspamd scores `SPOOF_REPLYTO(6.00) + FREEMAIL_REPLYTO_NEQ_FROM(2.00) = 8 pts` when `Reply-To` domain (yahoo.fr, gmail.com, etc.) differs from `From` domain (bewelcome.org). Combined with baseline spam signals this pushes past the Junk threshold. OTRS only polls INBOX, so submissions are silently lost.

We confirmed 4 emails from 2026-08-05 were stuck in `bw-otrs@bewelcome.org` Junk, including Gabriel's Safety and Software issues reports that triggered this investigation. Those 4 have been manually moved to INBOX via doveadm.

## Fix

Remove `Reply-To` from `sendFeedbackEmail` and `sendCommentReportedFeedbackEmail`. The reporter's email is already shown in the email body by both templates (the `feedback.html.twig` template was already updated in #471; `comment.feedback.html.twig` was not — fixed here by adding `{{ reporterEmail }}`).

OTRS agents can see and copy the reporter's address from the body. For OTRS workflow specifically, agents reply from within OTRS using the queue's system_address, not the raw email Reply button.

## What OTRS routing looks like

All feedback addresses (`abuse@`, `account@`, `bugs@`, etc.) are Mailcow aliases → `bw-otrs@bewelcome.org`. OTRS PostMaster filters match the original `To:` header and route to the correct queue automatically — no change needed there.

## Test plan

- [ ] Run `php bin/phpunit tests/Service/MailerTest.php` — all 3 tests pass
- [ ] Submit a test feedback form with a Yahoo/Gmail address — confirm it arrives in OTRS (not Junk)
- [ ] Report a comment as a user with a freemail address — confirm OTRS ticket is created

### DIFF
--- a/src/Service/Mailer.php
+++ b/src/Service/Mailer.php
@@ -93,6 +93,7 @@ class Mailer
     {
         $parameters['sender'] = $member;
         $parameters['receiver'] = $member;
+        $parameters['reporterEmail'] = $member->getEmail();
         $feedbackCategoryRepository = $this->entityManager->getRepository(FeedbackCategory::class);
         $feedbackCategory = $feedbackCategoryRepository->findOneBy(['name' => 'Comment_issue']);
 
@@ -101,7 +102,6 @@ class Mailer
             new Address($feedbackCategory->getEmailToNotify(), 'Comment Issue'),
             'comment.feedback',
             $parameters,
-            $member->getEmail(),
         );
     }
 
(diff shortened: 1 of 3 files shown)


Neophytis:
Closing in favour of an rspamd-side fix.

The root cause was that Mailcow's rspamd scored SPOOF_REPLYTO (+6) and FREEMAIL_REPLYTO_NEQ_FROM (+2) on feedback form emails where the Reply-To is the submitter's Gmail address. Combined with BAYES_SPAM and BAD_REP_POLICIES, the total score exceeded the spam-flag threshold, routing OTRS tickets to Junk.

The correct fix is two rspamd composites in composites.conf that cancel those symbols when R_DKIM_ALLOW and MAILCOW_DOMAIN_HEADER_FROM both fire — i.e. only for DKIM-authenticated mail from Mailcow-managed domains. This is gated on DKIM so a spammer cannot exploit it, and any spoofed noreply@bewelcome.org would still hit DMARC_POLICY_REJECT (+16) regardless.

Removing Reply-To at the application layer would break OTRS agent replies (agents would reply to noreply@bewelcome.org which discards mail instead of to the feedback submitter).

The fix is live on bobcat-prod. Score on feedback emails is now ~5.0 (no action), down from ~11 (add header).

PR #477, opened 2026-08-15
Feature/docker beta

Neophytis:
This is WIP please ignore and do not merge to master


docker/php/docker-entrypoint.sh
```diff
@@ -85,13 +85,16 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		if [ -f docker/db/geonamesadminunits.sql ]; then
 			mysql $database_name -u $database_user -p$database_password -h $database_host < docker/db/geonamesadminunits.sql
 		fi
-	elif ls -A Migrations/*.php > /dev/null 2>&1; then
+	elif ls -A migrations/*.php > /dev/null 2>&1; then
```
thisismeonmounteverest: Migrations on master are not doctrine migrations. Doctrine Migrations are stored in src/Migrations but are only used for the creation of the test database.

src/Service/Mailer.php
thisismeonmounteverest: I thought that was already addressed.

templates/emails/comment.feedback.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends 'emails/email.html.twig' %}
 
 {% block content%}
-    <p>The user {{ comment.toMember.username }} provided feedback for a comment:</p>
+    <p>The user {{ comment.toMember.username }} ({{ reporterEmail }}) provided feedback for a comment:</p>
```
thisismeonmounteverest: reporterEmail wouldn't be necessary here as comment.toMember.email would lead to the same.

tools/mvc-base/component.php
```diff
@@ -18,7 +18,7 @@ class RoxComponentBase
 
     public function __construct() {
         $this->setSession();
-        $this->entryPointLookup = new EntrypointLookup('build/entrypoints.json');
+        $this->entryPointLookup = new EntrypointLookup(__DIR__ . '/../../public/build/entrypoints.json');
```
thisismeonmounteverest: Why is this necessary? The code on www obviously finds the css and js files.

.trivyignore
thisismeonmounteverest: Wouldn't it be better to install that with the docker setup instead of ignoring it. The severity might change in the future.

composer.json
```diff
@@ -86,6 +85,7 @@
         "twig/intl-extra": "^3.0",
         "twig/markdown-extra": "^3.0",
         "twig/twig": "^3.0",
+        "willdurand/negotiation": "^3.1",
```
thisismeonmounteverest: Was wondering why this is suddenly needed, but was required as a dependency from api-platform before.

src/Command/ManticoreIndicesForumCommand.php
thisismeonmounteverest: These changes need to be removed.

src/Command/ManticoreIndicesGeonamesCommand.php
```diff
@@ -135,7 +135,7 @@ private function addGeonamesDocuments(Index $index, OutputInterface $output)
         do {
             $query = $this->entityManager->createNativeQuery(<<<___SQL
                 SELECT
-                    g.geonameid AS geonameid,
+                    g.geoname_id AS geonameid,
```
thisismeonmounteverest: The production database doesn't have a geoname_id column. This will break.

src/Controller/Admin/VolunteerToolController.php
thisismeonmounteverest: No geoname_id in geo__names.

Dockerfile
```diff
@@ -112,7 +118,7 @@ COPY pthacks pthacks/
 COPY public public/
 COPY roxlauncher roxlauncher/
 COPY src src/
-COPY Migrations Migrations/
+COPY migrations migrations/
```
thisismeonmounteverest: Unneccessary.

Dockerfile
```diff
@@ -25,7 +28,7 @@ RUN apk add --no-cache \
 		python3 \
 	;
 
-ARG APCU_VERSION=5.1.28
+ARG APCU_VERSION=5.1.18
```
thisismeonmounteverest: Why the downgrade? ACPu 5.1.22 is the first version compatible with PHP 8.2 and we're using 8.3 now.

docker-compose.yml
```diff
@@ -35,8 +29,6 @@ services:
     image: ${NGINX_IMAGE:-bewelcome/nginx}
     depends_on:
       - php
-    volumes:
-      - ./public:/srv/bewelcome/public:ro
 
   db:
     image: mariadb:12.0.2
```
thisismeonmounteverest: Use 12.3.2 instead.

thisismeonmounteverest:
Running ```docker-compose up -d``` to build a dev environment doesn't expose any port on the nginx image.

Neophytis:
> Running `docker-compose up -d` to build a dev environment doesn't expose any port on the nginx image.

please use branch neophytis/docker-infra-only and make all the modifications you want. This branch is for my testing purposes

Neophytis:
Closing — superseded by feature/docker-master.

PR #479, opened 2026-08-15
fix(migration): geonameId column name + widen Relations to VARCHAR(255)

Neophytis:
## Summary

Two bugs in the DB migration path, identified during the Lion→Alpha DB migration work and confirmed by Shevek's review.

**Bug 1 — `MigrateDatabaseCommand`: wrong column name for geonameId**

`migrateAddressTable()` was querying `g.geoname_id` (snake_case) but:
- The `Location` entity maps `@ORM\Column(name="geonameId")` — camelCase
- Lion's DB has the column as `geonameId`

The previous workaround was renaming the DB column in preflight, which would have broken Doctrine entirely. The fix is to query the column by its actual name.

**Bug 2 — `Version20260801120000`: `Relations` VARCHAR(87) too short**

The SET→VARCHAR migration used VARCHAR(87) for `comment.relations` and `comments.Relations`. Production data caused truncation. Widened to VARCHAR(255).

## Test plan
- [ ] Run `bin/console migrate:database` against a Lion DB copy — address ETL completes without "Unknown column 'g.geoname_id'" error
- [ ] Run `doctrine:migrations:migrate` — Version20260801120000 applies without truncation warnings


Neophytis:
Closing: these files (MigrateDatabaseCommand.php, Version20260801120000.php) exist only on develop, not on master or feature/docker-beta. Shevek owns the develop migration system. No changes from master perspective are needed here.

PR #483, opened 2026-08-20
ci(trivy): ignore CVE-2026-11940 until Alpine 3.24 backports python3 3.14.7-r0

Neophytis:
CVE-2026-11940 (HIGH, tarfile filter bypass in cpython) is failing the Trivy scan. The fix version 3.14.7-r0 exists in Alpine edge but is not yet backported to the Alpine 3.24 repo that the image uses. The Dockerfile already runs `apk upgrade` at build time, so this will resolve automatically once 3.24 ships the fix. Track: https://pkgs.alpinelinux.org/packages?name=python3&branch=v3.24


Neophytis:
Applied directly to feature/docker-beta.

PR #486, opened 2026-08-29
feat: Add DevContainer and Codespaces support for develop (FrankenPHP)

Ramjivan:
Following up on the discussion in #484, this PR adds the DevContainer setup for the \develop\ branch (FrankenPHP stack) so it can be tested in parallel with the \eature/docker-master\ stack (Nginx).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,61 @@
+{
+    "name": "BeWelcome Rox (Develop)",
+    "dockerComposeFile": [
+        "../docker-compose.yml",
+        "../docker-compose.override.yml.dist"
+    ],
+    "service": "php",
+    "workspaceFolder": "/srv/bewelcome",
+    "overrideCommand": false,
+    "forwardPorts": [80, 1080, 9306, 9308],
+    "portsAttributes": {
+        "80": {
+            "label": "BeWelcome App (FrankenPHP)",
+            "onAutoForward": "openBrowser"
+        },
+        "1080": {
+            "label": "MailCatcher Web UI",
+            "onAutoForward": "notify"
+        },
+        "9306": {
+            "label": "Manticore MySQL protocol",
+            "onAutoForward": "silent"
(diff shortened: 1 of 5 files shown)


Neophytis:
This PR is fine as a parallel local dev exploration of the FrankenPHP stack — xdebug is correctly placed in `bewelcome_php_dev` only, the Caddyfile is already on `develop`, and the `words.sql` typo fix is welcome.

That said: `develop` is not deployed anywhere. This is useful for contributors who want to experiment locally with FrankenPHP, but it should stay in draft until there is a deliberate team decision to migrate the production runtime. That decision needs more than a devcontainer — it needs full validation of the rox app under FrankenPHP, nginx removal, and sysadmins-infra changes.

No blocking issues in the diff itself.

Neophytis:
Closing — DevContainer support has been added via #484 targeting feature/docker-master. Thanks Ram!

Neophytis:
Closing this in favour of #484 which has already been merged into `feature/docker-master`. That PR covers the DevContainer setup we need for the Docker-based development environment. The FrankenPHP approach here is interesting but adds complexity we don't need at this stage — let's keep the stack consistent with what's running on the beta server. Thanks for the contribution Ram!

thisismeonmounteverest:
Reopened. We need to decide if we want to move to FrankenPHP in the future and if that the case the devcontainer is needed.

If we decide not to move the setup needs to be adapted to the one used in master.

The reason I choose FrankenPHP was that the docker stack is prepared for Mercury for push notifications. So I'd rather move.

Neophytis:
> Reopened. We need to decide if we want to move to FrankenPHP in the future and if that the case the devcontainer is needed.
> 
> If we decide not to move the setup needs to be adapted to the one used in master.
> 
> The reason I choose FrankenPHP was that the docker stack is prepared for Mercury for push notifications. So I'd rather move.

It's perfectly fine to evaluate it once Lion is retired. I had closed to not create confusion.

PR #503, opened 2026-09-19
fix(manticore): abort on existing index; optimize before translations phase

Neophytis:
## Summary

- `createGeonamesIndex()` and `createForumIndex()` had `$index = null` commented out in the catch block. When the table already existed, CREATE TABLE failed silently and the command returned the live index and inserted duplicate rows on every re-run. Uncommented the reset so it properly returns `null` and aborts.
- `addAlternateNamesDocuments()` crashes with an empty `ResponseException` after ~23 disk chunks accumulate during the 5.66M-row base phase. Added `OPTIMIZE INDEX geonames_rt OPTION cutoff=0, sync=1` between the two phases to merge all chunks to one before translations start.

## Current state on prod/stage

Both servers have `geonames_rt` with only `_geo` locale entries — translations were never indexed. Prod has 3× duplicates from repeated failed re-runs. After this merges, both indexes need a clean drop + rebuild:

```bash
docker exec manticore mysql -h127.0.0.1 -P9306 -e "DROP TABLE IF EXISTS geonames_rt; DROP TABLE IF EXISTS forum_rt;"
docker exec bewelcome-php-1 php bin/console manticore:indices:geonames
docker exec bewelcome-php-1 php bin/console manticore:indices:forum
```

## Test plan

- [ ] Merge triggers CI stage deploy
- [ ] Drop and rebuild on stage; confirm `SELECT COUNT(*), locale FROM geonames_rt GROUP BY locale` shows `_geo` + language rows (~10.4M total)
- [ ] Search "München" — should return Munich as top German result
- [ ] Same rebuild on prod after stage validates

Tracked in: BeWelcome/sysadmins-infra#543


thisismeonmounteverest:
From what I see happening on prod the final chunk of geonames throws an exception and after around half of the alternate names another exception is thrown. It really doesn't matter which chunk size is used.

Instead of checking if the index exists just drop the index at the point with no looking back.

